Issue #22, opened 2015-09-28
add docs directory to repo


pmuellr:
There's enough meaty information today collected in both the meetings and issues here, that I think it would be useful to create some documentation on the various components that have been discussed so far.  That would include:

* AsyncWrap
* v8-tracing
* platform-specific tracing (eg DTrace, LTTng, ETW for Windows)
* perf-events (from Yunong)

I would expect this to be rather fluid, as we've not settled on any stories for anything yet.  So I'd expect lots of updates and evolution.

We could also do this in the wiki, but I find the wiki a bit harder to deal with.

Qard:
:+1: The wiki feature is not great. Putting this stuff in the repo would probably be better.

PR #26, opened 2015-09-29
initial stab at some tracing docs

pmuellr:
for issue #22 "add docs directory to repo"

### DIFF
--- a/docs/AsyncWrap/README.md
+++ b/docs/AsyncWrap/README.md
@@ -1,0 +1,16 @@
+Node.js tracing - AsyncWrap
+================================================================================
+
+AsyncWrap is the central location in Node.js that handles async things
+specific to Node.js.  As such, it's a likely place where tracing will be
+desired.  Should serve as a good customer of the more general tracing story for
+Node.js.
+
+An older-ish intro to AsyncWrap is available in Trevor Norris's blog post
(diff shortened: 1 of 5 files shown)


Qard:
LGTM

mikeal:
This is awesome!

Is there anywhere that we are exposing these other than here? I'd like to get these viewable from the website.

Qard:
I can help getting it into the docs repo. What are your thoughts on the progress docs state of some of this though? The trace_event.h stuff is not really usable at all yet, so I'm not sure what the value of having that on the website would be.

What *might* be interesting is to give working groups a more public-facing space to share progress and inform people what work needs to be done. People could just go digging in the github repos, but that doesn't feel quite as welcoming.

AndreasMadsen:
AsyncWrap is specifically left out of the official documentation so that it doesn't need to obey semver.

Qard:
Yeah, I definitely wouldn't want to put this stuff in API docs; this is all use-at-your-own-risk material. It could use more visibility though.

pmuellr:
@mikeal I think at this point the docs aren't really designed to be public-facing; we have lots of info collected over minutes, blog posts, other conversations, and these docs are intended to be a single place to collect that stuff.

For now.

Eventually, when/if anything becomes public-facing-able, totally makes sense to add to the official docs.

PR #27, opened 2015-10-05
docs: add up to date documentation for AsyncWrap

AndreasMadsen:
This should be an up to date documentation on the AsyncWrap API. I have not included in discussion on the future of the API, but there a small section (Things you might not expect) on some stuff that other users should know.

Also I'm not really sure what the purpose of the provider argument is and how it compares to `this.constructor.name` someone else should elaborate on that. 

PS: I've heard my spelling isn't ideal, bear with me.

### DIFF
--- a/docs/AsyncWrap/README.md
+++ b/docs/AsyncWrap/README.md
@@ -25,7 +25,7 @@ handle objects.
 The API that node exposes is mostly defined in JavaScript files. However
 ECMAScript does not define any API for creating a TCP socket, reading a file
 etc.. That logic is implemented in C++ using libuv and the v8 API. The JavaScript in
-nodecore interacts with this C++ layer using the handle objects.
+node core interacts with this C++ layer using the handle objects.
 
 For example in `net.connect(port, address)` that creates a TCP connection,
 two handle objects (`TCPConnectWrap` and `TCP`) are created:
@@ -49,7 +49,7 @@ one (`TCP`) is for maintaining the connection.
 
 `TCPConnectWrap` gets its information by setting properties on the handle
 object, like `address` and `port`. Those properties are read by the C++ layer,
-but can also be inspected from the AsyncWrap hooks. When the handle is create
+but can also be inspected from the AsyncWrap hooks. When the handle is created
 using `new TCPConnectWrap()` the `init` hook is called.
 
 A `oncomplete` property is also set, this is the callback for when the
@@ -58,8 +58,8 @@ is called, just after the `after` hook is called.
 
 The `TCP` handle works exactly the same way, except that the information
 is passed as arguments to a method `.connect` and the `onread` function
-is called multiply times, thus it behaves like an event. This also means that
-the `before` and `after` hooks are called multiply times.
+is called multiple times, thus it behaves like an event. This also means that
+the `before` and `after` hooks are called multiple times.
 
 Thus one should expect the hooks be called in the following order:
 
@@ -84,7 +84,8 @@ the text above and the text below.
 
 ## The API
 
-At the moment there don't exist a high level API for AsyncWrap. It is simply exposed through `process.binding`:
+At the moment there is not a high level API for AsyncWrap. It is simply exposed
+through `process.binding`:
 
 ```javascript
 const asyncWrap = process.binding('async_wrap');
@@ -135,9 +136,9 @@ signatures are quite similar. The `this` variable refers to the handle object,
 and `init` hook has two extra arguments `provider` and `parent`.
 
 ```javascript
-function init(provider, parent) { this = handle; }
-function before() { this = handle; }
-function after() { this = handle; }
+function init(provider, parent) { }
+function before() {  }
+function after() { }
 ```
 
 ##### this
@@ -256,11 +257,11 @@ lookup before that.
 inside one of the hooks, creates an infinite recursion. Use `fs.syncWrite(1, msg)`
 or `process._rawDebug(msg)` instead.
 
-* `process.nextTick` never creates a handle object. You will have to money patch
+* `process.nextTick` never creates a handle object. You will have to monkey patch
 this.
 
 * Timer functions (like `setTimeout`) shares a single `Timer` handle, thus you
-will usually have to money patch those functions.
+will usually have to monkey patch those functions.
 
 ## Resources
 
--- a/docs/AsyncWrap/README.md
+++ b/docs/AsyncWrap/README.md
@@ -27,7 +27,8 @@ ECMAScript does not define any API for creating a TCP socket, reading a file
 etc.. That logic is implemented in C++ using libuv and the v8 API. The JavaScript in
 nodecore interacts with this C++ layer using the handle objects.
 
-For example to create and connect a TCP socket:
+For example in `net.connect(port, address)` that creates a TCP connection,
+two handle objects (`TCPConnectWrap` and `TCP`) are created:
 
 ```javascript
 const TCP = process.binding('tcp_wrap').TCP;
@@ -43,15 +44,16 @@ socket.onread = onread;
 socket.connect(req, address, port);
 ```
 
-This example uses two handle objects. The first one (`TCPConnectWrap`) is for
-connecting the socket, the second one (`TCP`) is for maintaining the connection.
+The first one (`TCPConnectWrap`) is for connecting the socket, the second
+one (`TCP`) is for maintaining the connection.
 
 `TCPConnectWrap` gets its information by setting properties on the handle
 object, like `address` and `port`. Those properties are read by the C++ layer,
 but can also be inspected from the AsyncWrap hooks. When the handle is create
 using `new TCPConnectWrap()` the `init` hook is called.
 
-A `oncomplete` property is also set, this is the callback for when the connection is made or failed. Just before calling `oncomplete` the `before` hook
+A `oncomplete` property is also set, this is the callback for when the
+connection is made or failed. Just before calling `oncomplete` the `before` hook
 is called, just after the `after` hook is called.
 
 The `TCP` handle works exactly the same way, except that the information
@@ -77,7 +79,8 @@ after // TCP
 ...
 ```
 
-_tick_ indicates there is at least one tick between the text above and the text below.
+_tick_ indicates there is at least one tick (as in `process.nextTick()`) between
+the text above and the text below.
 
 ## The API
 
@@ -94,7 +97,7 @@ To assign the hooks call:
 
 ```javascript
 asyncWrap.setupHooks(init, before, after);
-function init(provider) { /* consumer code */ }
+function init(provider, parent) { /* consumer code */ }
 function before() { /* consumer code */ }
 function after() { /* consumer code */ }
 ```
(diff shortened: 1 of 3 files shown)


pmuellr:
@trevnorris can you review?

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,202 @@
...
+before // TCP
+after // TCP
+=== tick ===
+before // TCP
+after // TCP
+=== tick ===
+...
+```
+
+_tick_ indicates there is at least one tick between the text above and the text below.
```
trevnorris: The word "tick" is unfortunately ambiguous. Do you mean in the sense of `nextTick()` or as in a tick of the event loop?

One thing that isn't clear from this is which operation generated which object. For example. `net.createServer()` will create a `TCP`, and `net.connect()` will create the `TCPConnectWrap`, but another `TCP` is created that is the actual connection to the server. This will have it's own init callback called.

Part of this behavior is pending on https://github.com/nodejs/node/pull/3216, which operates more like I had intended it to from the beginning. Log from the callbacks of that test to see what I mean.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,202 @@
...
+function after() { /* consumer code */ }
+```
+
+Note that calling `asyncWrap.setupHooks` again, will overwrite the previous
+hooks.
+
+#### Enable and disable
+
+Because there is a small performance penalty in just calling a noop function,
+AsyncWrap is not enabled by default. To enable AsyncWrap call:
```
trevnorris: Side note, I haven't allowed noop functions because I figured they wouldn't really be useful. Though it would be easy enough with my latest patch to allow each to be optional. Thoughts? (also, probably for discussion elsewhere)

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,202 @@
...
+```
+
+#### The hooks
+
+Currently there are 3 hooks: `init`, `before` and `after`. The function
+signatures are quite similar. In all cases the `this` variable is the handle
+object. Users may read properties from this object such as `port` and `address` in the `TCPConnectWrap` case, or set user specific properties.
+
+Finally the `init` has an extra argument called `provider`, this is an
+integer that refer to names defined in an `async.Providers` object map.
```
trevnorris: Should make note that is isn't safe to iterate the `this` in the `init` callback. The object is partially constructed and there is at least one getter that'll segfault because the property isn't yet available. Though it is safe to assign values and read known properties.

Also note that with my previously mentioned patch the second argument in the init callback will be the parent object. Since it's impossible for all children to be created within the before/after callback scope, it can be used to perform better tracing.

trevnorris:
@AndreasMadsen Thanks for all the text. I think my referenced patch is the last for the near future. There are some grammatical things, but we'll worry about those later.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,202 @@
...
+
+* `console.log` is async and thus invokes AsyncWrap, thus using `console.log`
+inside one of the hooks, creates an infinite recursion. Use `fs.syncWrite(1, msg)`
+or `process._rawDebug(msg)` instead.
+
+* `process.nextTick` never creates a handle object. You will have to money patch
+this.
+
+* Timer functions (like `setTimeout`) shares a single Timer handle, thus you
+will usually have to money patch those functions.
```
trevnorris: Another thing to add would be that because of some quirks in the node API, disabling at certain times can lead to unexpected outcomes. For example:
```js
async_wrap.enable();
let server = net.createServer();
async_wrap.disable();
server.listen(0);
```
In the above the server will not be tracked. Because the `TCPWrap` is not instantiated until `listen` is called. Reason for this is because we don't know whether a PIPE or TCP will be used until that time. It's been very frustrating for me, but not something I've been able to easily work around.

trevnorris: Additionally the following:
```js
async_wrap.enable();
let server = net.createServer().listen(0, 'localhost');
async_wrap.disable();
```
will not track the TCP server because the first call `listen()` makes is to dns to resolve `'localhost'`. This may be a shortcoming of the current API, as the way `parent` propagates could be made more generic, but for the time being it doesn't work.

docs/AsyncWrap/README.md
AndreasMadsen: * I mean in the `nextTick()` sense.
* It is `net.connect()` that created both `TCPConnectWrap` and `TCP`.

Here is a script that gives the shown output: https://gist.github.com/AndreasMadsen/a90591087368fd42496b (well it only does one onread call).

docs/AsyncWrap/README.md
AndreasMadsen: > Should make note that is isn't safe to iterate the this in the init callback. 

Good point. I actually had this exact issue, and it took a long time to debug :)

> the second argument in the init callback will be the parent object

I can see it has landed, I will update the documentation.

AndreasMadsen:
@trevnorris Thanks for the review. I have updated the documentation based on your comments and https://github.com/nodejs/node/pull/3216 (add `parent` object).

docs/AsyncWrap/README.md
AndreasMadsen: I don't consider it particularly important at this stage.

watson: I still think that the call to `enable()` should be required, but when setting up the hooks I think there are cases where optional functions are nice if that's what you're asking. I've had situations where I was logging some stuff related to init and didn't care about before/after - it was worse though back when the error function was still a part of this... never used that one :wink:

trevnorris: > I still think that the call to enable() should be required

Elaborate?

As for making them optional, that's a trivial patch at this point because of one of the most recent patches.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+```
+
+#### The hooks
+
+Currently there are 3 hooks: `init`, `before` and `after`. The function
+signatures are quite similar. The `this` variable refers to the handle object,
+and `init` hook has two extra arguments `provider` and `parent`.
+
+```javascript
+function init(provider, parent) { this = handle; }
```
Qard: The assignment of `this = handle` feels a little wrong to me...

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+## Handle objects
+
+AsyncWrap emits events (hooks) that informs the consumer about the life of all handle
+objects in node. Thus in order to understand AsyncWrap one must first understand
+handle objects.
+
+The API that node exposes is mostly defined in JavaScript files. However
+ECMAScript does not define any API for creating a TCP socket, reading a file
+etc.. That logic is implemented in C++ using libuv and the v8 API. The JavaScript in
+nodecore interacts with this C++ layer using the handle objects.
```
Qard: `nodecore` is not a proper word. Maybe put a space in there?

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+socket.onread = onread;
+socket.connect(req, address, port);
+```
+
+The first one (`TCPConnectWrap`) is for connecting the socket, the second
+one (`TCP`) is for maintaining the connection.
+
+`TCPConnectWrap` gets its information by setting properties on the handle
+object, like `address` and `port`. Those properties are read by the C++ layer,
+but can also be inspected from the AsyncWrap hooks. When the handle is create
```
Qard: *created

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+using `new TCPConnectWrap()` the `init` hook is called.
+
+A `oncomplete` property is also set, this is the callback for when the
+connection is made or failed. Just before calling `oncomplete` the `before` hook
+is called, just after the `after` hook is called.
+
+The `TCP` handle works exactly the same way, except that the information
+is passed as arguments to a method `.connect` and the `onread` function
+is called multiply times, thus it behaves like an event. This also means that
```
Qard: *multiple

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+the `before` and `after` hooks are called multiply times.
```
Qard: *multiple

docs/AsyncWrap/README.md
AndreasMadsen: Yes, I felt that too. Do you have a suggestion?

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+=== tick ===
+...
+```
+
+_tick_ indicates there is at least one tick (as in `process.nextTick()`) between
+the text above and the text below.
+
+## The API
+
+At the moment there don't exist a high level API for AsyncWrap. It is simply exposed through `process.binding`:
```
Qard: s/don't exist/is not/

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,275 @@
...
+
+* `console.log` is async and thus invokes AsyncWrap, thus using `console.log`
+inside one of the hooks, creates an infinite recursion. Use `fs.syncWrite(1, msg)`
+or `process._rawDebug(msg)` instead.
+
+* `process.nextTick` never creates a handle object. You will have to money patch
+this.
+
+* Timer functions (like `setTimeout`) shares a single `Timer` handle, thus you
+will usually have to money patch those functions.
```
Qard: *monkey

docs/AsyncWrap/README.md
Qard: I think it's fine to leave the function bodies empty. The rest of the text in that section adequately explains that `this` is the handle.

AndreasMadsen:
Thanks @Qard, I have updated the documentation.

pmuellr:
I assume we're ready to merge this?  Or are more updates coming?

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
 Node.js tracing - AsyncWrap
 ================================================================================
 
-AsyncWrap is the central location in Node.js that handles async things
-specific to Node.js.  As such, it's a likely place where tracing will be
-desired.  Should serve as a good customer of the more general tracing story for
-Node.js.
+AsyncWrap is two things. One is a
+[class abstraction](https://github.com/nodejs/node/blob/master/src/async-wrap.h)
+there provides node internal facilities for handling async things, such as
```
trevnorris: "that provides an internal mechanism for handling asynchronous tasks, such as..."

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+calling a callback.
+The other part is a an API for setting up hooks, that allow one to get
```
trevnorris: "is an API for setting up hooks and allows one to get..."

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+structural tracing information about the life of handle objects. In the context
+of tracing the latter is usually what is meant.
 
-An older-ish intro to AsyncWrap is available in Trevor Norris's blog post
-[AsyncWrap Tutorial - An Introduction](http://blog.trevnorris.com/2015/02/asyncwrap-tutorial-introduction.html).
+_The reasoning for the current naming confusion, is that the API part implements
```
trevnorris: remove comma

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+the hooks through the AsyncWrap class. But there is nothing inherently correct
+in doing that, for example if v8 provided those facilities the AsyncWrap class
```
trevnorris: "the AsyncWrap class, but this is not inherently necessary. For example..."

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+would not need be involved in the AsyncWrap API._
 
-There was also some discussion in [issue #21](https://github.com/nodejs/tracing-wg/issues/21#issuecomment-142727693).
+For the remaining description the API part is what is meant by AsyncWrap.
+
+## Handle objects
```
trevnorris: Capitalize "Objects". Do similar for other headers in document.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+
+AsyncWrap emits events (hooks) that informs the consumer about the life of all handle
```
trevnorris: "that inform"

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+objects in node. Thus in order to understand AsyncWrap one must first understand
+handle objects.
+
+The API that node exposes is mostly defined in JavaScript files. However
+ECMAScript does not define any API for creating a TCP socket, reading a file
```
trevnorris: Maybe something like "Node's core API is mostly defined in JavaScript. However..."

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+
+## Things you might not expect
+
+* It is not obvious when a handle object is created. For example the TCP server
+creates the `TCP` handle when `.listen` is called and it may perform an DNS
+lookup before that.
+
+* `console.log` is async and thus invokes AsyncWrap, thus using `console.log`
+inside one of the hooks, creates an infinite recursion. Use `fs.syncWrite(1, msg)`
+or `process._rawDebug(msg)` instead.
```
trevnorris: possibly mention that `fs.syncWrite()` doesn't call `util.inspect()` so the output will be unformatted? Not a big thing.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,276 @@
...
+
+* `process.nextTick` never creates a handle object. You will have to monkey patch
+this.
+
+* Timer functions (like `setTimeout`) shares a single `Timer` handle, thus you
+will usually have to monkey patch those functions.
+
```
trevnorris: Possibly add a note about Promises also aren't tracked.

trevnorris:
finished another round. mainly grammer and stuff. great stuff.

AndreasMadsen:
Thanks @trevnorris.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,281 @@
...
+## The API
+
+At the moment there is not a high level API for AsyncWrap. It is simply exposed
+through `process.binding`:
+
+```javascript
+const asyncWrap = process.binding('async_wrap');
+```
+
+_Be warned that this API is not an official API and can change at any time, even if it's just patch
```
trevnorris: Long line.

docs/AsyncWrap/README.md
```diff
@@ -1,15 +1,281 @@
...
+}
+
+function getStack(message) {
+  const localStack = new Error(message);
+  return localStack.stack + '\n' + currentStack;
+}
+module.exports = getStack;
+```
+
+Please note that this example is way simpler, than what is required from a
```
trevnorris: remove comma.

trevnorris:
Two nits. Other than that I would say LGTM. Nice bit of writing you did!

AndreasMadsen:
Fixed.

AndreasMadsen:
@Qard @pmuellr Any last comments.

Qard:
LGTM

PR #37, opened 2015-12-28
Async wrap update - destroy hook and uid

AndreasMadsen:
This documents the changes made in https://github.com/nodejs/node/pull/3461

/cc @trevnorris

### DIFF
--- a/docs/AsyncWrap/README.md
+++ b/docs/AsyncWrap/README.md
@@ -53,13 +53,13 @@ but can also be inspected from the AsyncWrap hooks. When the handle is created
 using `new TCPConnectWrap()` the `init` hook is called.
 
 A `oncomplete` property is also set, this is the callback for when the
-connection is made or failed. Just before calling `oncomplete` the `before` hook
-is called, just after the `after` hook is called.
+connection is made or failed. Just before calling `oncomplete` the `pre` hook
+is called, just after the `post` hook is called.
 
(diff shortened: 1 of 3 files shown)


docs/AsyncWrap/README.md
```diff
@@ -131,25 +152,33 @@ asyncWrap.disable();
 
 #### The Hooks
 
-Currently there are 3 hooks: `init`, `before` and `after`. The function
-signatures are quite similar. The `this` variable refers to the handle object,
-and `init` hook has two extra arguments `provider` and `parent`.
+Currently there are 4 hooks: `init`, `pre`, `post` `destroy`. The `this`
+variable refers to the handle object. The `init` hook has three extra arguments
```
trevnorris: May want to make a note that the associated JS object can usually be accessed via `this.owner`. This isn't official, but is the common practice. Maybe we want to solidify this? /cc @bnoordhuis

docs/AsyncWrap/README.md
```diff
@@ -131,25 +152,33 @@ asyncWrap.disable();
...
+`provider`, `uid` and `parent`. The `destroy` hook also has the `uid` argument.
```
trevnorris: gah. deficiency in the API. will make it so the `uid` is passed to pre/post.

trevnorris:
Thanks for the docs update. LGTM.

PR #45, opened 2016-04-04
docs: update after nodejs/node#5756

AndreasMadsen:
nodejs/node#5756 changed `setupHooks` to take an object as input and added `didThrow` as an argument to the `post` hook.

PR #48, opened 2016-04-27
setupHooks pre node v6 / starting with node v6

baslr:
as of https://github.com/nodejs/node/blob/a20c70029075cab18e4ff4d27dfd95b7eda2b268/test/parallel/test-async-wrap-uid.js#L9

starting with node version 6, you have to pass an object


AndreasMadsen:
Thanks. There have been a few more changes. I made a pull request a while ago which covers them all, https://github.com/nodejs/tracing-wg/pull/45/files

AndreasMadsen:
Thanks for looking at #45

PR #59, opened 2016-06-15
Updated info on Inspector Protocol (CDP)

joshgav:
This adds
- an initial list of informational resources about the Inspector Protocol
- tables to track support in tools and runtimes for protocol domains.

Is this the right format to track compat?
Can tool and runtime owners update their rows?

/cc @nodejs/diagnostics @ofrobots @auchenberg @jasonLaster @develar

### DIFF
--- a/debugging/README.md
+++ b/debugging/README.md
@@ -1,29 +1,86 @@
-## Debugging API
-- [Chrome Debugging Protocol](https://chromedevtools.github.io/debugger-protocol-viewer/)
-- [V8 Debugging Protocol](https://github.com/v8/v8/wiki/Debugging-Protocol)
+## Debugging API and Protocol
 
-Node currently relies on V8's internal Debug API and associated commands and events. The Debug API is published at [v8/include/v8-debug.h](https://github.com/v8/v8/blob/master/include/v8-debug.h), and the message protocol is documented in prose in [the V8 wiki](https://github.com/v8/v8/wiki/Debugging-Protocol) and in code in [v8/src/debug/debug.js#L2341-2372](https://github.com/v8/v8/blob/master/src/debug/debug.js#L2341-L2372). 
+### Inspector Protocol
+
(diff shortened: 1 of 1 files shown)


auchenberg:
I've started on the RemoteDebug Compatibility tables that compares `protocol.json`'s from Chrome, Safari, Edge and Node. 

It's early days, but it should provide a better overview of what parts of the API that are stable and implemented across runtimes: http://compatiblity.remotedebug.org/

pavelfeldman:
@joshgav: Should we rather track the clients of the protocol for Node along with their capabilities? I am not sure listing the backend methods various backends expose makes much sense for Node - that should be clients' business to declare / track the backends they support.

develar:
IntelliJ IDEs support ("use") `Debugger`, `Console`, `Runtime`, `Worker` (doesn't make any sense for node, right?).

joshgav:
@pavelfeldman 

> I am not sure listing the backend methods various backends expose makes much sense for Node

Is this because you believe Node should provide a single set of domains regardless of backend VM? I was thinking such a standard set would _only be a subset_ of the capabilities of V8 or other VMs, and the full set could be documented here.

In the short term V8 is the only VM and it's reasonable to focus mostly on it. But I hope our work here will contribute to JavaScript diagnostics in the long term, so wanted to try to keep other runtimes in mind.

Going back to edit this now. Will add @auchenberg's links too. Thanks!

pavelfeldman:
> But I hope our work here will contribute to JavaScript diagnostics in the long term, so wanted to try to keep other runtimes in mind.

Not sure I follow the long term goal here. What exactly are you trying to achieve?

joshgav:
@pavelfeldman trying to keep in mind that our effort here is the likely foundation for diagnostic interfaces for non-browser JS in the future, including future Node and Node alternatives.

But agree that Node+V8 are primary.

joshgav:
Closing.

Ongoing protocol discussion can continue in @auchenberg's links above and perhaps at https://github.com/ChromeDevTools/debugger-protocol-viewer.

Debugging getting started guide for web site proposed here: https://github.com/nodejs/nodejs.org/pull/1131

PR #87, opened 2017-02-16
wg: nominating jkrems

joshgav:
Nominating @jkrems to the Diagnostics WG. Jan is the original author of [node-inspect](https://github.com/nodejs/node-inspect) and as a member could better contribute to that project.

@jkrems this does depend on you wanting to be a member of the WG. The main requirement is that you check and contribute to issues here occasionally and try to attend meetings. Your work on node-inspect is already more than enough IMO.

We should discuss at the next meeting according to protocol. To be sure we have consensus, anyone from @nodejs/diagnostics who has objections should please chime in here. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,3 +46,4 @@ Work needed includes:
 * @danielkhan
 * @Fishrock123
 * @naugtur
+* @jkrems


joshgav:
Discussed at meeting and no objections! Thanks for joining us @jkrems!

jkrems:
I think for CI purposes it would be important to be added to some Github team as well (unless that already happened..?), [as per](https://github.com/nodejs/node-inspect/pull/25#issuecomment-280451745):

> And specifically, he does not need to be a collaborator, just a member of the team configured for the diagnostics workgroup.

joshgav:
@jkrems 

> it would be important to be added to some Github team as well

Indeed, now we need someone to add you to the [@nodejs/diagnostics](https://github.com/orgs/nodejs/teams/diagnostics) team. I'm not an admin there; @Fishrock123 @mhdawson @ofrobots can one of you add @jkrems to that team, and perhaps make me an admin at the same time so I can do that in the future? Thank you!

mhdawson:
I've triggered the invite to jkrems to have him added to the team.

mhdawson:
@joshgav I promoted you to a team maintainer as I think that makes sense given your involvement.

jkrems:
Worked, thanks! :)

PR #105, opened 2017-09-22
Removing matthewloring from diagnostics wg

matthewloring:
I will be moving on to a new position and will have less time
going forward to work on Node.js diagnostics.


AndreasMadsen:
Good luck :)

Qard:
I should probably remove myself too at some point, given that I've had just about zero involvement for over a year. 😅

PR #132, opened 2017-12-13
Remove outdated AsyncWrap docs

watson:
Since the AsyncHooks docs are now online, I can't think of a reason for the Diagnostics WG to maintain this documentation any more. As discussed [in todays meeting](https://github.com/nodejs/diagnostics/issues/115) I'm therefore making this PR to remove the old AsyncWrap docs from the repo.

The official docs can be found here:
https://nodejs.org/api/async_hooks.html

@AndreasMadsen: This PR is just a suggestion. Just let me know if you think we should do something else 😃


AndreasMadsen:
Hehe, agreed. But we should update https://github.com/nodejs/diagnostics/blob/master/tracing/README.md too.

watson:
@AndreasMadsen should I update (and rename) the AsyncWrap section in the [tracing README.md](https://github.com/nodejs/diagnostics/blob/master/tracing/README.md) or do you want me to just remove it completely?

I think it would be nice to have something still as it's a module that falls under our area. I can try and update this PR with a suggestion

watson:
@AndreasMadsen For starters, I just made a minimal change to the tracing README.md now. I'm not sure if the actual text needs changing as well

AndreasMadsen:
Maybe just:

> [AsyncHooks](https://nodejs.org/api/async_hooks.html) - Tracks the lifetime of asynconuse requests (init/destroy) and their callback activity (before/after).

We now track more than just the `AsyncWrap` C++ classes. In particular we allow node modules to use the Embedder API to make pools look correct (like setTimeout does).

watson:
@AndreasMadsen fixed

PR #133, opened 2017-12-15
Add list of modules breaking async continuity

watson:
Just starting off the list some some well known sinners 😄 

Not sure if the phrase "async continuity" makes sense to people, but it was the best I could come up with - feel free to suggest something else

### DIFF
--- a/tracing/AsyncHooks/problematic-modules.md
+++ b/tracing/AsyncHooks/problematic-modules.md
@@ -1,0 +1,18 @@
+# User-land modules breaking async continuity
+
+This is a list of user-land Node modules that breaks async continuity.
+The list is not meant to be complete, but feel free to make a PR if you
+feel anything is missing.
+
+Over time we'll hopefully work together with the module owners to
+implement the [Async Hooks Embedder
+API](https://nodejs.org/api/async_hooks.html#async_hooks_javascript_embedder_api)
+so that these modules no longer needs to be monkey patched. But even if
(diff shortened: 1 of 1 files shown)


jkrems:
Should we link to existing PRs like https://github.com/petkaantonov/bluebird/pull/1472?

watson:
@jkrems good idea - didn't know there were already a PR. I've updated the table with a link to it.

jkrems:
Good to merge?

p-bakker:
Don't think generic-pool should be on this list: it uses global.Promise by default (so whatever you have installed) and allows to provide the provide the Promise implementation to use through configuration

ofrobots:
It seems that several APMs are monkey-patching `generic-pool` to propagate context:
* Google Cloud's [Stackdriver](https://github.com/googleapis/cloud-trace-nodejs/blob/fae65bd86d3444f2e50254a7f3fe55a55460af8d/src/plugins/plugin-generic-pool.ts).
* Elastic Node.js [agent](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/generic-pool.js)
* Probably more.

From looking at the patches, it seems that at least `acquire` loses context. If that is still accurate, `generic-pool` belongs on the list.

Qard:
It's mostly because of pre-promise versions of generic-pool, but also because `global.Promise` could be set to _any_ promise implementation, not just the native one, so it _could_ be one not supported by our instrumentation.

PR #183, opened 2018-04-18
Adding minutes from 2018-04-18 meeting

mike-kaufman:


wg-meetings/2018-04-18.md
```diff
@@ -0,0 +1,74 @@
...
+   - no other progress
+ - Visualizing Async Context [#179](https://github.com/nodejs/diagnostics/issues/179)
+  
+ - Diag WG Deep Dives - topics [#168](https://github.com/nodejs/diagnostics/issues/168)
+  - please update if anyone has any topics 
+
+ - Node CPU Profiling Roadmap [#148](https://github.com/nodejs/diagnostics/issues/148)
+   - PR in v8 was merged last week
+   - will be available in v8 6.7 release
+   - will node v10 ship with 6.6 or 6.7?  Unknown
```
mmarchini: v10.0.0 will ship with 6.6 and it's going to be updated to 6.7 eventually (https://github.com/nodejs/node/issues/19354#issuecomment-373157341).

wg-meetings/2018-04-18.md
```diff
@@ -0,0 +1,74 @@
...
+   - Ali:  is this worth backporting to v8 6.6?
+     - maybe, no discussion yet
```
mmarchini: @hashseed doesn't feel optimistic about backporting. I could open a backport CL or floating PR for this though and see how it goes. @ofrobots WDYT?

mmarchini: In both cases, it wouldn't land cleanly and we would need to make some manual changes to make it work (Node.is 8 is probably worse though). 

Some things I can think we would need to do:
* There's no GetSharedFuncionCode for builtins and CSA prior to V8 6.7, and part of the implementation is done inside this function
* We would need to find all calls to `SharedFunctionInfo::bytecode_array` and change it to `SharedFunctionInfo::BytecodeArray`
* We would need to find all calls to `SharedFunctionInfo::set_bytecode_array` and make it behave correctly when the flag is on
 
I'm probably missing something. I don't mind making a float PR, but I'm concerned it would be instantly rejected with so many changes.

mmarchini: Tried to apply the patch to master and I got conflicts in 18 files. Even if we decide to move forward with this, I'm worried this might make subsequent floating patches harder to apply, since it changes the name of `SharedFunctionInfo::bytecode_array` to `SharedFunctionInfo::BytecodeArray`. Maybe we'll just need to accept that this will only land on Node.js 10 and only after it gets V8 6.7.

One thing we might do to minimize conflict with subsequent floating patches is to keep `SharedFunctionInfo::bytecode_array` in snake_case. Would need some major changes in the patch, but at least it would generate less conflicts and wouldn't get in the way of new patches.

image

PR #197, opened 2018-05-24
Adding doc describing "async context" in terms of "continuations"

mike-kaufman:
See [here](https://github.com/mike-kaufman/diagnostics/tree/mkaufman-add-async-context/async-context) for a rendered view of the markdown.

This is a refinement of the work we've been doing in http://github.com/mike-kaufman/async-context-definition/.  I would really appreciate any feedback people have on the terminology, concepts, correctness, or the flow of the document - my intent here is this is easily explainable to a broad section of JS programmers.

### DIFF
--- a/async-context/README.md
+++ b/async-context/README.md
@@ -12,10 +12,10 @@ We strive to adhere to some basic principles in this document.  We believe that 
 3.  Constructs should enhance JavaScript programmers' understanding of async code flow & async boundaries.
 4.  Model should allow for simplified reasoning about async code execution.
 5.  Resulting Data Structures and APIs and should support solving common "async understanding" use cases (e.g., long stack traces, continuation local storage, visualizations, async error propogation, user-space queueing...).
-6.  Model should allow for implementations at different layers of the stack (VM, host or "monkey-patched"), and implementations in appropriate languages (c++ or JavaScript).
+6.  It should be straightforward to implement this model at any API layer (i.e., VM, host or "monkey-patched").
 
 ## A Simple Example
-We'll start with a simple example that is not controversial. Most JS programmers should have an intuitive understanding of where the async boundaries are in this code.  We'll build off of this example going forward.  Implicit in this example is an undefined concept of "async context".  We'll build stronger definitions later on, but for now, we'll use "async context" to refer to JS programmers' colloquial understanding of asynchonrous function calls.
+We'll start with a simple example that is not controversial: 
 
 ```javascript
 function f1() {
@@ -29,7 +29,11 @@ function f1() {
         }
     }, 1000);
 }
+
+f1();
 ```
+
+Most JS programmers should have an intuitive understanding of what happens when the code above is executed.  We'll build off of this example going forward.  Implicit in this example is an undefined concept of "async context".  We'll build stronger definitions later on, but for now, we'll use "async context" to refer to JS programmers' colloquial understanding of asynchonrous function calls.
 
 There are three interesting constructs in the above example:
   - First, the function `f2` is a function created in one "async context" and passed through an API.  When `f2` is invoked later, it is a "logical continuation" of the "context" in which it was created.
@@ -112,7 +116,7 @@ and
 
 Some notes about the pictures above:
   - We've introduced a "Root Continuation" as the bottom frame.  This  illustrates a basic assumption that all code is executing in the context of a `Continuation Invocation`.
-  - Multiple `continuations` can be on the stack at the same time, which result in mulitple `Continuation Invocations`. For any given stack frame, the "context" is the first `Continuation Invocation` below the given frame on the stack.
+  - Multiple `continuations` can be on the stack at the same time, which result in mulitple `Continuation Invocations`.
 
 Here, we've updated our pictures of runtime callstacks, giving  labels to the `Continuation Invocation` instances:
 
@@ -148,11 +152,25 @@ and
 ===================================            -----------
 ```
 
+## Continuation Invocation States
+A `Continuation Invocation` can be in a number of states:
+  - `pending` - A  `Continuation Invocation` instance has been created, but is not yet executing, nor is it ready to execute.
+  - `cancelled` - A `Continuation Invocation` was abandoned.
+  - `ready` - A `Continuation Invocation` is ready to execute, but is not yet executing
+  - `executing` - `Continuation Invocation` is currently executing on the stack
+  - `paused` - A `Continuation Invocation` has started execution, but is not currently on the stack.  E.g., for generator functions. // todo - is this necessary for async/await & generators.  what is tc39 terminology?
+  - `completed` - A `Continuation Invocation` instance is completed.  However, the instance, during execution, may have called other `Continuation Points` and those `Continuation Invocations` are still pending execution.
+  - `collectable` - A `Continuation Invocation` instance is completed and all child-spawned continuations are in the collectable state.
+
+## Context and Current Context
+For any given stack frame, we define it's  **Context** as the `Continuation Invocation` defined by the first `Continuation` below the given frame on the stack.  For example, in the previous diagram, the stack frame of `function x()` has a `Current Context` of `ci3`,
+and the stack frame of  `... host code...` has a "`Current Context` of `ci2`.  We define the **Current Context** as the `Context` of the top frame on the stack. 
+
 ## Link Context
-We define the `Link Context` `link-c` of a `Continuation` `c` as a pointer from `c` to the `Continuation Invocation` instance where `c` was constructed.  In the example from above, the `continuation` `f2` has a `link context` pointing to `Continuation Invocation` `ci1`.
+We define the **Link Context** of a `Continuation` to be the `Current Context` when a `Continuation` is constructed.   In the example from above, the `continuation` `f2` has a `link context` pointing to `Continuation Invocation` `ci1`.
 
 ## Ready Context
-Consider the following example:
+We define the **Ready Context** to be the `Current Context` when a `Continuation Invocation` changes state from `pending` to `ready`.  This is useful for promises, where we want to understand the series of `Continuation Invocation` instances on a promise chain.  When a promise is resolved or rejected, the promise implementation must modify the state of `Continuation Invocations` associated with subsequent promises on the chain. Consider the following example:
 
 ```javascript
 const p = new Promise((resolve, reject) => {
@@ -174,9 +192,7 @@ In this code example, we have:
     the initial code is executing in. 
   - four `Continuation Invocations`.
 
-We define the `Ready Context` (//TODO, close on naming - `Causal Context`, `Resolve Context`???) for a promise as a reference from a `Continuation Invocation` associated with a Promise.then call, to the `Continuation Invocation` in which the preceding promise in the promise chain was resolved.
-
-For non-promise-based APIs, the `Ready Context` is the same as the `Link Context` (//TODO or should it be null/undefined?)
+For non-promise-based APIs, the `Ready Context` is the same as the `Link Context`.
 
 In our example above, the `Continuation Invocation` associated with `continuation f2()` has as its `Ready Context`, a reference to the `Continuation Invocation` associated with `continuation f3()`.  
 
@@ -189,7 +205,14 @@ In our example above, the `Continuation Invocation` associated with `continuatio
   - **Continuation** -  a function that, when invoked, creates a new `Continuation Invocation` instance, and maintains references to other related `Continuation Invocation` instances. Using TypeScript for descriptive purposes, a `Continuation` has the following shape:
 
     ```TypeScript
-    interface Continuation {
+    /**
+     * maker interface for a generic function type
+     */
+     interface IFunction {
+        (...args: any[]): any
+    }
+
+    interface Continuation extends IFunction{
         linkingContext: ContinuationInvocation;
     }
     ```
@@ -204,16 +227,7 @@ In our example above, the `Continuation Invocation` associated with `continuatio
     }
     ```
 
-  - **Async Call Graph** - A directed acyclic graph comprised of `Continuations` and `Continuation Invocations` instances as nodes, and the `linkingContext` and `readyContext` references as edges.
-
-### Continuation Invocation States
-A `Continuation Invocation` can be in a number of states:
-  - `pending` - A  `Continuation Invocation` instance has been created, but is not yet executing, nor is it ready to execute.
-  - `ready` - A `Continuation Invocation` is ready to execute, but is not yet executing
-  - `executing` - `Continuation Invocation` is currently executing on the stack
-  - `paused` - A `Continuation Invocation` has started execution, but is not currently on the stack.  E.g., for generator functions. // todo - is this necessary for async/await & generators.  what is tc39 terminology?
-  - `completed` - A `Continuation Invocation` instance is completed.  However, the instance, during execution, may have called other `Continuation Points` and those `Continuation Invocations` are still pending execution.
-  - `collectable` - A `Continuation Invocation` instance is completed and all child-spawned continuations are in the collectable state.
+  - **Async Call Graph** - A directed acyclic graph comprised of `Continuations` and `Continuation Invocations` instances as nodes, and the `linkingContext` and `readyContext` references as edges. // TODO draw some pictures of the DAG
 
 ## Where are the Continuation Points?
 The `Continuation Points` are defined by convention by the host environment.  The host needs logic to determine if a given argument is a function or a continuation, and if not yet a continuation, needs to "continuify" the parameter.  The host is also responsible for "pinning" any `Continuation` instances to prevent premature garbage collection.
@@ -258,6 +272,7 @@ The events above are JSON events and their schema is illustrated in the followin
 // TODO: add better examples here. 
 
 ## Applications
+Most applications can be thought of traversal of a specific path of edges through the `Async Call Graph`.  We give some trivial examples below.  
 
 ### Continuation Local Storage
 ```typescript
--- a/async-context/README.md
+++ b/async-context/README.md
@@ -2,7 +2,9 @@
 # JavaScript Asynchronous Context
 
 ## Overview
-This document defines a model for "JavaScript Asynchronous Context".  This model includes terminology & definitions, programmatic "types" that result from these definitions, and relations between these types.  These utlimately form an API that expose the model, and allow us to reason about and solve common JavaScript "async code execution" problems
+This document defines a model for "JavaScript Asynchronous Context".  This model includes terminology & definitions, programmatic "types" that result from these definitions, and relations between these types.  These utlimately form an API that expose the model, and allow us to reason about and solve common JavaScript "async code execution" problems.
+
+We frame the problem in two parts.  First is a directed acyclic graph we call the "Async Call Graph" that accurately models the relationships of asynchronous code execution.  With a well-defined, common problems can be defined in terms of traversals over this graph.
 
 ## Principles
 We strive to adhere to some basic principles in this document.  We believe that adhereing to these principles will ensure simplicity and help guide the model.
@@ -18,6 +20,9 @@ We strive to adhere to some basic principles in this document.  We believe that 
 We'll start with a simple example that is not controversial: 
 
 ```javascript
+function x(s) { ... }
+function myLoggingAPI(s) { x(s); }
+
 function f1() {
     let i = 0;
     let interval = setInterval(function f2() {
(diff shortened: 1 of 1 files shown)


mike-kaufman:
//cc @nodejs/diagnostics, @mrkmarron, @ofrobots, @kjin

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+|    ...host code...      |                              |
+---------------------------                              \/
+|    ...host code...      |                   Continuation Invocation ci2                   
+===================================                      /\
+|   Host Continuation B()         |                      |
+===================================            -----------
+```
+
+## Link Context
+We define the `Link Context` `link-c` of a `Continuation` `c` as a pointer from `c` to the `Continuation Invocation` instance where `c` was constructed.  In the example from above, the `continuation` `f2` has a `link context` pointing to `Continuation Invocation` `ci1`.
```
kjin: It's difficult for me to parse this sentence. I'm not sure how to improve it, but it seems necessary to first mention that there exists some parent/child relationship between continuations, and that the `pointer` refers to an edge illustrating that relationship.

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+});
+```
+
+In this code example, we have:
+  - `setTimeout` and `Promise.then` as our `continuation points`. 
+  - four `continuations` - `f1`, `f2` and `f3`, plus the "root continuation" that 
+    the initial code is executing in. 
+  - four `Continuation Invocations`.
+
+We define the `Ready Context` (//TODO, close on naming - `Causal Context`, `Resolve Context`???) for a promise as a reference from a `Continuation Invocation` associated with a Promise.then call, to the `Continuation Invocation` in which the preceding promise in the promise chain was resolved.
```
kjin: `Ready Context ` seems like the best of the three.

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+For non-promise-based APIs, the `Ready Context` is the same as the `Link Context` (//TODO or should it be null/undefined?)
```
kjin: My two cents -- there is no need for it to be `null/undefined`.

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+// TODO:  add stacks w/ `continuation invocations` to illustrate `ready context`.
+
+## Crisper Definitions
+
+  - **Continuation Point** - a function that accepts a `Continuation` as a parameter.  This logically represents an async boundary; functions passed to a `Continuation Point` are invoked at some later point in time.
+
+  - **Continuation** -  a function that, when invoked, creates a new `Continuation Invocation` instance, and maintains references to other related `Continuation Invocation` instances. Using TypeScript for descriptive purposes, a `Continuation` has the following shape:
+
+    ```TypeScript
+    interface Continuation {
```
kjin: Is it something like `Continuation extends Function`?

mike-kaufman:
@kjin - thanks for comments, all makes sense. I'm going off the grid until next wed - will incorporate your feedback then.

async-context/README.md
mike-kaufman: was thinking that the "ready context" can be generalized to the current context when specific state changes happen on a `Continuation Invocation`.  This seems straightforward, and it's a nice definition independent of promises.

async-context/README.md
ofrobots: > where `c` was constructed

I think 'where `c` was passed to a continuation point is more apt. Consider:

```js
const f1 = function() {...}
... somwhere else ...
fs.readFile(filename, f1);
```

littledan:
cc @domenic

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+---------------------------
+|   setTimeout()          |
+---------------------------
+|   function f1()         |
+---------------------------
+|    ...host code...      |
+---------------------------
+|    ...host code...      |
+---------------------------
+|    host function A()    |
```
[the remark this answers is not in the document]
mike-kaufman: My opinion is that distracts from the conversation with unnecessary detail.  Here's the actual stack:

```
f1 (e:\temp\async-example\async-example.js:3)
(anonymous function) (e:\temp\async-example\async-example.js:13)
Module._compile (loader.js:699)
Module._extensions..js (loader.js:713)
Module.load (loader.js:612)
tryModuleLoad (loader.js:551)
Module._load (loader.js:543)
Module.runMain (loader.js:744)
startup (node.js:238)
bootstrapNodeJSCore (node.js:572)
```

Would love to hear what others think here.

async-context/README.md
mike-kaufman: That was my first thought, but I kinda feel like "where constructed" is 
  - easy to implement
  - has very clear semantics
  - lets us push some anti-patterns back on the user as a "don't do that".  

Note that per current definitions in this doc, a `Link Context` is associated with the `Continuation` and not the `Continuation Invocation`.  Furthermore, a single `Continuation` can be associated with multiple `Continuation Invocations`.

Now, consider this example where we have two APIs that are `continuation points`, e.g., `fs.read()` and `stream.read()`, and `fs.readFile()` calls `stream.read()`:

```javascript
function readFile(filename, callback) {
   const stream = openFileStream(filename);
   stream.read(callback);
}
```

In the example above, if we define link context as "where a continuation is passed to a continuation point", where is the link context above established?  Whereas, if we define it as where the `continuation` is created, then it becomes very obvious.

In your example, we can push back on the user and say "don't do that". 

Happy to discuss this further.

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Principles
+We strive to adhere to some basic principles in this document.  We believe that adhereing to these principles will ensure simplicity and help guide the model.
+
+1.  Definitions should flow from well-understood, well-defined constructs in synchronous programming.
+2.  Definitions should be independent of the host environment (e.g., Node.js, the browser).
+3.  Constructs should enhance JavaScript programmers' understanding of async code flow & async boundaries.
+4.  Model should allow for simplified reasoning about async code execution.
+5.  Resulting Data Structures and APIs and should support solving common "async understanding" use cases (e.g., long stack traces, continuation local storage, visualizations, async error propogation, user-space queueing...).
+6.  Model should allow for implementations at different layers of the stack (VM, host or "monkey-patched"), and implementations in appropriate languages (c++ or JavaScript).
```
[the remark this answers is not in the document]
mike-kaufman: reworded.

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+    let interval = setInterval(function f2() {
+        if ( i < 2) {
+            myLoggingAPI(`i is ${i}`);
+            ++i;
+        } else {
+            clearInterval(interval);
+        }
+    }, 1000);
+}
+```
```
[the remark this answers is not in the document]
mike-kaufman: done

async-context/README.md
```diff
@@ -0,0 +1,360 @@
...
+2.  Definitions should be independent of the host environment (e.g., Node.js, the browser).
+3.  Constructs should enhance JavaScript programmers' understanding of async code flow & async boundaries.
+4.  Model should allow for simplified reasoning about async code execution.
+5.  Resulting Data Structures and APIs and should support solving common "async understanding" use cases (e.g., long stack traces, continuation local storage, visualizations, async error propogation, user-space queueing...).
+6.  Model should allow for implementations at different layers of the stack (VM, host or "monkey-patched"), and implementations in appropriate languages (c++ or JavaScript).
+
+## A Simple Example
+We'll start with a simple example that is not controversial. Most JS programmers should have an intuitive understanding of where the async boundaries are in this code.  We'll build off of this example going forward.  Implicit in this example is an undefined concept of "async context".  We'll build stronger definitions later on, but for now, we'll use "async context" to refer to JS programmers' colloquial understanding of asynchonrous function calls.
```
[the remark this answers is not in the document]
mike-kaufman: done

async-context/README.md
mike-kaufman: tried to re-word.  LMK if it helps.

async-context/README.md
mike-kaufman: changed this to be context when the Continuation Invocation's state changes to ready.

async-context/README.md
mike-kaufman: updated.

mcollina:
Very good work.

In https://github.com/mike-kaufman/diagnostics/blob/11206a47094846bc0b1027ab82c7bd0fa4f04d87/async-context/README.md#continuation-local-storage I would use a private symbol to store the cls.

I think the example there of walking the "linking context" is very expensive. Moreover, I'm not necessarily convinced it works in the case of await. Consider the following:

```js
async function f2 () {
  await something()
  // consider this point B
}

async function f1 () {
  await f2()

  // consider this point A
} 
```

The problem with walking the context during get() (as in the example in the document), is that whatever we add in point B _will not be included_ in point A. I believe the users would expect the two continuations to be linked, but they are not. 

This is because of how async await is specified (I'm not the best person to explain why, cc @bmeurer), and the promise f2() and the result of await are parallel continuations. 

I do not think it is possible to handle CLS without a `init`-like hook to manually propagate the context instead of walking the linked contexts.

vdeturckheim:
Great stuff.

Can we define what is what I call a "continuation loss" meaning for some reason `ContinuationInvocation.GetCurrent()` ends up returning something wrong?

mike-kaufman:
>Can we define ... a "continuation loss" meaning for some reason ContinuationInvocation.GetCurrent() ends up returning something wrong?

@vdeturckheim  - can you provide some examples of where/why this would happen.  Ideally, this is pathological behavior.  i.e., we provide clear & crisp definitions, and the implementation returns "the right thing" based on these definitions.  This is the "Async Call Graph".   

Now, it's possible that different applications want to traverse the graph in different ways, but I wouldn't call this a "continuation loss".

async-context/README.md
```diff
@@ -0,0 +1,375 @@
...
+   - A **Continuation Point** - is a function that accepts a **Continuation** as a parameter.  This logically represents an async boundary; functions passed to a **Continuation Point** are invoked at some later point in time.  In our example above, `setInterval` is a **Continuation Point**. 
+  -  A **Continuation Invocation** is a specific invocation of a `continuation`; more precisely, a **Continuation Invocation** is the *period of time* where a `continuation` frame is on the stack.  As soon as the continuation frame pops off the stack, then that **Continuation Invocation** is completed. Note that a `continuation` instance can have more than one **Continuation Invocation** instances associated with it.  In our code sample above, there are precisely two **Continuation Invocation** instances associated with the `continuation` f2. 
+
+## A lower-level view
+
+At runtime we can view the example above as two distinct call stacks at specific points in program execution.  These look something like this:
+
+```
+---------------------------
+|   setTimeout()          |
```
Flarna: I think this should be `setInterval()`.

async-context/README.md
```diff
@@ -0,0 +1,375 @@
...
+---------------------------
+|    host function B()     |
+---------------------------
+```
+
+Since we previously made a distinction between a regular JavaScript function and a `continuation`, let's make the same distinction in our pictures of the callstacks:
+
+```
+--------------------------- 
+|   setTimeout()          |
```
Flarna: I think this should be `setInterval()`.

vdeturckheim:
@mike-kaufman I agree this is a case that should not happen. 
However, due to implementation details, it is not always possible to find a parent to a node of the graph (I had the issue with an hand-crafted promise library recently).

async-context/README.md
```diff
@@ -0,0 +1,375 @@
...
+===================================            -----------
+```
+
+## Continuation Invocation States
+A `Continuation Invocation` can be in a number of states:
+  - `pending` - A  `Continuation Invocation` instance has been created, but is not yet executing, nor is it ready to execute.
+  - `cancelled` - A `Continuation Invocation` was abandoned.
+  - `ready` - A `Continuation Invocation` is ready to execute, but is not yet executing
+  - `executing` - `Continuation Invocation` is currently executing on the stack
+  - `paused` - A `Continuation Invocation` has started execution, but is not currently on the stack.  E.g., for generator functions. // todo - is this necessary for async/await & generators.  what is tc39 terminology?
```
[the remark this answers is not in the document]
mike-kaufman: @Flarna - I think so.  I'll review implementation in Chakra & the spec, but I believe Generator functions yield & restore the call stack, so by definitions above, the generator invocation will be in a single `continuation invocation`.  Again, I need to work through more examples with generators & async/await, so if you have an example where this doesn't make sense, please let me know.

async-context/README.md
```diff
@@ -0,0 +1,375 @@
...
+});
+```
+
+In this code example, we have:
+  - `setTimeout` and `Promise.then` as our `continuation points`. 
+  - four `continuations` - `f1`, `f2` and `f3`, plus the "root continuation" that 
+    the initial code is executing in. 
+  - four `Continuation Invocations`.
+
+For non-promise-based APIs, the `Ready Context` is the same as the `Link Context`.
```
Flarna: Is this also true for server like objects where functions get called because of some external event (e.g. incoming TCP connection on a net server,...)?

mike-kaufman:
@mcollina  - thanks for the feedback.  Please keep comments like this coming! 

Sorry for length of post below.  Happy to get on phone to clarify anything.

1.  RE "very expensive", I'm not following this, can you clarify?  My take here is operations like I've described are equivalent to walking a list, and these are O(n) where n is the length of the "path-to-the-root" from the current node.  I'd like to keep the "cost analysis" observations concrete - either runtime/memory big-O complexity of the algorithms being proposed, or grounded in real repeatable measurements. Please LMK if I'm missing something.

1.  RE the async/await example, I think it is helpful here to not conflate the structure of the graph with the algorithms over top of it.  That is, I'd like to arrive at a clear set of definitions & rules, and we can apply those in a rigorous way, and the result is a data structure that accurately represents the "async call graph".   Once we agree on these rules & the graph structure, then we can break higher-level problems down into "traversal" problems - i.e., does traversing a specific set of edges give us desired behavior?  I think your comment is really a traversal problem in two parts, namely:
  i. what are user expectations of "CLS" in async-await examples?
  ii. how do we traverse the graph to achieve user-expected resutls?

I'll look into the implementation of generators and async/await in more detail to see check my assumptions here.  For now, I'll break your example down into it's functionally equivalent promise-based implementation, since I think that makes the async boundaries more clear:

```javascript
function f2 () {
  return somethingThatReturnsAPromise()
    .then(function f3() { // line 3
       // consider this point B
   });
}

function f1 () {
  return f2()
      .then(function f4() {  // line 10
         // consider this point A
  });
} 
```

Given definitions above, we have the following stack at the time that `somethingThatReturnsAPromise()` is called:


```
----------------------------------                ----------- 
| somethingThatReturnsAPromise() |                          |
----------------------------------                          \/                   
|    function f2()               |                  Continuation Invocation ci1
----------------------------------                          /\    
|    function f1()               |                          |
----------------------------------                          |
|    ...host code...             |                          |
======================================                      |
|   Root Continuation A()            |                      |
======================================            -----------

```

And we have the following stack at the time `then` is called on "line 3":

```
----------------------------------                ----------- 
|    promise.then() // line 3    |                          |
----------------------------------                          \/                   
|    function f2()               |                  Continuation Invocation ci1
----------------------------------                          /\    
|    function f1()               |                          |
----------------------------------                          |
|    ...host code...             |                          |
======================================                      |
|   Root Continuation A()            |                      |
======================================            -----------

```

And we have the following stack at the time `then` is called on "line 10":

```
----------------------------------                ----------- 
|    promise.then() // line 10   |                          |
----------------------------------                          \/
|    function f1()               |                  Continuation Invocation ci1
----------------------------------                          /\    
|    ...host code...             |                          |
======================================                      |
|   Root Continuation A()            |                      |
======================================            -----------

```

Note that all of the above are executing within the same `Continuation Invocation`, namely `ci1`.  Now, when the `continuations` `f3` and `f4` are invoked, they each have their own `continuation invocation` instance - let's call these `ci-f3` and `ci-f4` respectively.  The (simplified) `Async Call Graph` looks like this:


```
            ------- 
            | ci1 |
            -------
               /\
               |
   ---------------------------------
   |                               |
   |                               |
 "link edge"                   "link edge"
   |                               |
---------                       --------- 
| ci-f3 |<----"ready-edge"------| ci-f4 |
---------                       ---------

```


Now, to address  your questions:

1.  Is it reasonable for a user to expect a CLS value stored in `ci-f3` to be accessible in `ci-f4`?  I don't claim to know the answer here, but I would argue that these are peer branches of the async call tree, thus such a value would not be "visible". Would love to hear other opinions/arguments here.

2.  If you wanted to make a value stored in `ci-f3` accessible in `ci-f4`, can you do that?  I claim the answer here is "yes", namely by traversing `cif-f4`'s "ready-edge" on lookup.

I've been thinking a bit about how dynamically prescribe traversal paths through the graph for specific functionality.  (e.g., how do we choose the "correct path" through the graph to do CLS"?)  I think the answer here is to allow dynamic labels over the edges such that a "CLS iterator" can choose the path of edges labeled "CLS" as it traverses.  This needs to be communicated in more detail, but I think it nicely accomodates the expected behavior you suggest.

async-context/README.md
mike-kaufman: :) fixed.

async-context/README.md
mike-kaufman: fixed

async-context/README.md
mike-kaufman: I'd like to get to a stronger definition here, and I'm thinking of defining the `X Context` to be the `Current Context` when the `Continuation Invocation` instance changes to state `X`.  I like this because it gives the host some flexibility to account for host-specific things in a way that remains consistent with the definitions.  And, it generalizes the "Ready Context" away from promises. Let me know if you have thoughts on this.

async-context/README.md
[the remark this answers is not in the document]
Flarna: I'm also not sure here. Thinking of  `await` I would wonder that a single `Continuation Invocation` for following async function is correct:
```
async function foo() {
  // A
  await some();
  // B
  await thing();
  // C
  await else();
  // D
}
```
Here I would expect that each parts A, B, C and D have their own invocation and that they are chained.

Flarna:
I think besides linking continuations we have to think also about combining them - and decide which link context is the right one (or both?).
Think about following:
```
async function foo() {
  const p1 = doFoo();
  const p2 = doBar();
  return Promise.all([p1, p2]);
}
await foo();
```
Here we have two resolvers for the returned promise. In this sample the link context of `p1` and `p2` are the same but it should be not that hard to create a sample where there are different.

async-context/README.md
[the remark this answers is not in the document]
mike-kaufman: > Here I would expect that each parts A, B, C and D have their own invocation and that they are chained

@Flarna - This is my intuition as well.  See comment [here](https://github.com/nodejs/diagnostics/pull/197#issuecomment-394458286).  Certainly the case if you look at above example as simple promises instead of async/await, e.g.,: 

```javascript
async function foo() {
  // A
  return some().then(() => {
      // B
      return thing();
  }).then(()=> {
     // C
     return else();
  }).then(() => {
     // D   
  })
}
```

I'm going to dig into the async/await details later this week & I'll update then if any of this changes.

async-context/README.md
```diff
@@ -0,0 +1,375 @@
+
+# JavaScript Asynchronous Context
+
+## Overview
+This document defines a model for "JavaScript Asynchronous Context".  This model includes terminology & definitions, programmatic "types" that result from these definitions, and relations between these types.  These utlimately form an API that expose the model, and allow us to reason about and solve common JavaScript "async code execution" problems
+
```
mhdawson: nit. Missing period at end of line.

async-context/README.md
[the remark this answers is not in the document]
mhdawson: * I found this a bit confusing as well to start.  Maybe we need a definition of 'Host X' to clarify when we say Host followed by Continuation or function (see second comment) that these are frames added by the host which are not part of the program itself.

* I'd change `... host code ...` to  `...host function()...` as I think that is more consistent with the rest of what is on they stacks as shown.

* It might be useful to define some of the `standard` host continuations as I think it would have been a lot more obvious if the second stack had started with `Host Continuation Timeout()` and the first with `Host Continuation Main()`

async-context/README.md
```diff
@@ -0,0 +1,375 @@
...
+---------------------------
+|    host function A()    |
+---------------------------
+```
+
+and
+
+```
+---------------------------
+| function x()            |
```
mhdawson: I think if you are going to add x(), its probably a good idea to include in the example myLoggingAPI() showing it calling x().

Otherwise maybe just remove x() from the stack?

async-context/README.md
mike-kaufman: fixed

async-context/README.md
mike-kaufman: fixed for clarity

gireeshpunathil:
can this be landed? I skimmed through this an d looks like lot of useful information, but not sure how updated it is, given its age. @mhdawson @Qard - can you pls have a look and suggest?

mhdawson:
@gireeshpunathil I think we might as well land it and then we can update if needed.

Qard:
Maybe merge and add a note at the top that the doc needs editing?

RafaelGSS:
> Maybe merge and add a note at the top that the doc needs editing?

+1

gireeshpunathil (approved):

gireeshpunathil:
landed in bdcfe0ea33e2a63f72def5eeac29a02aac80f403

Issue #217, opened 2018-08-02
Membership review - still active?


mhdawson:
Membership review as agreed in: https://github.com/nodejs/diagnostics/issues/215

@AndreasMadsen - Andreas Madsen -> Active
@bnoordhuis - Ben Noordhuis
@brycebaril - Bryce Baril -> emeritus
@cjihrig - Colin Ihrig -> Active
@danielkhan - Daniel Khan -> Actrive
@dberesford - Damian Beresford
@Fishrock123 - Jeremiah Senkpiel
@Flarna - Gerhard Stöbich ->active
@hashseed - Yang Guo ->active
@hekike - Peter Marton ->active
@Hollerberg - Gernot Reisinger -> Active
@jasnell - James M Snell ->active
@jkrems - Jan Olaf Krems -> Active
@joyeecheung - Joyee Cheung ->active
@kjin - Kelvin Jin -> Active
@mcollina - Matteo Collina -> active
@mhdawson - Michael Dawson ->active
@mike-kaufman - Mike Kaufman -> active
@mmarchini - Matheus Marchini ->active
@naugtur - Zbyszek Tenerowicz ->active
@ofrobots - Ali Ijaz Sheikh -> active
@othiym23 - Forrest L Norvell -> active
@piscisaureus - Bert Belder -> emeritus
@pmuellr - Patrick Mueller -> emeritus
@Qard - Stephen Belanger -> active
@rmg - Ryan Graham -> emeritus
@sam-github - Sam Roberts -> active
@thekemkid - Glen Keane
@thlorenz - Thorsten Lorenz -> emeritus
@trevnorris - Trevor Norris
@watson - Thomas Watson -> active
@yunong - Yunong Xiao -> active

Please confirm by posting a comment on this issue if you plan to continue being active in the Diagnostics WG. If we don't hear from you we will move you to emeritus as outlined in #215 .

mhdawson:
Have already heard from (in issue discussion review in the first place)

@rmg  -> emeritus
@naugtur -> active
@hekike  -> active
@jasnell ->active
@hashseed ->active
@ofrobots ->active
@yunong ->active
@mmarchini ->active
@Flarna ->active

joyeecheung:
I am still active

AndreasMadsen:
I'm still here if you need me :)

Hollerberg:
still breathing

cjihrig:
Also still breathing. Sometimes

kjin:
[...]

PR #224, opened 2018-08-25
resync team with README

Trott:
@rmg removed themselves from the @nodejs/diagnostics team on June 20. Resynch the list of people in the README against the GitHub team.

(The team currently has 31 people. If there are really that many people active in the Diagnostics WG, then awesome! But if that contains a lot of folks who aren't active anymore, it might be a good idea to do some pruning. Or not if extra folks don't adversely affect anything. In some groups, that can be a problem when doing things like taking votes or abiding by rules about quorum.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,6 @@ Work needed includes:
 - [@piscisaureus](https://github.com/piscisaureus) - Bert Belder
 - [@pmuellr](https://github.com/pmuellr) - Patrick Mueller
 - [@Qard](https://github.com/Qard) - Stephen Belanger
-- [@rmg](https://github.com/rmg) - Ryan Graham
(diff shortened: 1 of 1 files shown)


richardlau:
> (The team currently has 31 people. If there are really that many people active in the Diagnostics WG, then awesome! But if that contains a lot of folks who aren't active anymore, it might be a good idea to do some pruning. Or not if extra folks don't adversely affect anything. In some groups, that can be a problem when doing things like taking votes or abiding by rules about quorum.)

There's a revalidation effort currently going on in https://github.com/nodejs/diagnostics/issues/217.

Trott:
(Can someone on Diagnostics land this?)

Issue #107, opened 2017-10-07
Async-context formalization and diagnostics support


mrkmarron:
Updating text here to better reflect reality - *mike-kaufman*

**Current Status:**
 - [x] high-level description of async context is checked into Diag repo in form of [slides](https://github.com/nodejs/diagnostics/tree/master/async-context/slide-decks)
 - [ ] Update  slides to account for the following
    - [ ]  explicit APIs for how to do property lookups.  Support "lazy init promises" 
         ```javascript
         var p;
         app.get('/', (req, res) = {
            if (!p) { 
                p = someAsyncFuncThatReturnsAPromise(); 
            }
            p.then(() => {…});
         });
         ```
    - [ ]  Promise.all(), Promise.race()
    - [ ]  Async/Await examples
    - [ ]  generator examples
    - [ ]  async generator examples
    - [ ]  async recurision and path compression support (e.g., `f = () = > { setTimeout(f, 1000); }`)
    - [ ]  specific API changes (I.e., what APIs/properties get exposed through JS to users)
 - [ ]  Get plan on how to take to TC-39
 - [ ]  Identify how ready-context can be implemented based on current promise hooks APIs
 - [ ] Understand perf impact

jkrems:
> **causal context:** relates f^i causes g^j if during the ex- ecution of f^i
> the function g^j is enabled for execution via an E-Then rule application
> on an already resolved priority promise or via an E-Resolve rule application
> on a previously un-resolved priority promise.
>
> [...]
>
> For programs that do not use JavaScript promises and rely on raw callbacks the linking context and causal context will be the same at all program points.

I assume the above is meant as "without additional signals" / "in the scope of the formal language described in the paper"? E.g. the following doesn't use "real" Promises but still de-facto has different linking/causal context if I understand it correctly:

```js
var connection;
var queued = [];

function runWithConnection(cb) { // userland E-Then
  if (connection) { cb(connection); }
  else { queued.push(cb); }
}

function setConnection(c) { // userland E-Resolve
  connection = c;
  queued.forEach(f => f(c));
}

setConnection(net.connect(/* ... */)); // ~ calling E-Resolve
runWithConnection(postConnect); // ~ calling E-Then
```

jkrems:
> Investigate how the general formalisms in the paper handle various corner cases (such as Promise.all).

Would it make sense to collect these in a runnable form? E.g. starting an acceptance suite that just defines the cases but not even necessarily the expected values for causal and/or linking context?

[...]

PR #247, opened 2018-10-17
adding minutes from 2018-10-17 WG meeting

mike-kaufman:


wg-meetings/2018-10-17.md
```diff
@@ -0,0 +1,72 @@
...
+         - Some discussion of reviving async resource.  Goal is to
+           remove the destroy hook, or so optional that it does not interfere with performance, PR
+           to remove promise from Promise wrap.  Will make escape analysis easier but requires
+          work on the V8 side to use escape analysis in emitting the hook
+    - Ali need Yang and Matteo to go into more details. Ali will schedule a separate time
+    for the discussion.
+  
+  - Async-context formalization and diagnostics support [#107](https://github.com/nodejs/diagnostics/issues/107)
+    - will schedule a deep-dive for november
+    - Natalie & Bryan from new relic are going to do some code
```
mmarchini: Is this in a public repo? I would like to help :)

mike-kaufman: @mmarchini  - We'd love any help you  can provide here.  A few places to dig in:
   -  There's a number of //todos now listed in #107. 
   -  @NatalieWolf started a Node.JS implementation:  https://github.com/NatalieWolfe/node-context-continuation.   I think this could benefit from playing with it, as well as filling in the details around the ready context hook-up.
  - I'm really interested in what this would look like when wired up at the VM level, so I think there's opportunity to do this as a strawman, in V8 and/or Chakra-Core.  If Chakra-Core, I can help with guidance on what needs to be written (I haven't found the time...).

nataliewolf: Tagging the correct Natalie Wolf so they can see this :) @nataliewolfe

PR #286, opened 2019-03-22
doc: Remove `Node report in core` initiative

gireeshpunathil:
Remove `Node report in core` from Current list of
initiatives, as this is achieved. Report has landed.

Refs: https://github.com/nodejs/node/pull/22712

cc @mhdawson


gireeshpunathil:
landed as 4deda165197a352c42214b9ee56f6d9b19c97a15

PR #343, opened 2019-12-13
Add heap snapshot and gc traces howtos

naugtur:
Requesting feedback on comments below.


documentation/memory/step2/using_heap_snapshot.md
```diff
@@ -6,4 +6,37 @@ can also compare multiple snapshots to see differences over time.
...
+1. Continue using the functionality for a while, preferably without running anything else in between.
+1. Take another heap snapshot. The difference between the two should mostly contain what was leaking.
+1. Open Chromium/Chrome dev tools and go to *Memory* tab
+1. Load the older snapshot file first, newer one second ![Screenshot: Load button in tools]
+(TODO) <!--not sure if we want to maintain the screenshots-->
+1. Select the newer snapshot and switch mode in a dropdown at the top from *Summary* to *Comparison*.
+1. Look for large positive deltas and explore the references that caused them in the bottom panel.
+
+
+Practice capturing heap snapshots and finding memory leaks with [a heap snapshot exercise](https://github.com/naugtur/node-example-heapdump)
```
naugtur: This is a bit more than a case-study and I don't see how it'd fit into the case_study.md but it serves the same purpose.

documentation/memory/step3/using_gc_traces.md
```diff
@@ -1 +1,23 @@
 //TODO
+
```
naugtur: Didn't remove the TODO as this still feels like a stub. 
What else would go into this file?

naugtur:
What's the policy on screenshots? 
They would be useful here, but someone would have to update them if devtools change.

documentation/memory/step2/using_heap_snapshot.md
```diff
@@ -6,4 +6,37 @@ can also compare multiple snapshots to see differences over time.
...
+If you need a snapshot form a working process, like an application running on a server, you should implement getting it using:
+
+```js
+require('v8').writeHeapSnapshot()
+```
+
+Check [writeHeapSnapshot docs](https://nodejs.org/dist/latest-v12.x/docs/api/v8.html#v8_v8_writeheapsnapshot_filename) for file name options
+
+You need to have a way to invoke it without stopping the process, so calling it in a http handler or as a reaction to a signal from the operating system is advised.  
+Be careful not to expose the http endpoint triggering a snapshot. It should not be possible for anybody else to oaccess it.
```
legendecas:
```suggestion
Be careful not to expose the http endpoint triggering a snapshot. It should not be possible for anybody else to access it.
```

documentation/memory/step2/using_heap_snapshot.md
```diff
@@ -6,4 +6,37 @@ can also compare multiple snapshots to see differences over time.
 
 ## How To
 
-//TODO
+### Get the Heap Snapshot
+
+The simplest way to get a Heap Snapshot is to connect a debugger to your process running locally and go to Profiling, choose to take a memory snapshot.
+
+If you need a snapshot form a working process, like an application running on a server, you should implement getting it using:
```
gireeshpunathil:
```suggestion
If you need a snapshot from a working process, like an application running on a server, you should implement getting it using:
```

documentation/memory/step2/using_heap_snapshot.md
```diff
@@ -6,4 +6,37 @@ can also compare multiple snapshots to see differences over time.
...
+
+```js
+require('v8').writeHeapSnapshot()
+```
+
+Check [writeHeapSnapshot docs](https://nodejs.org/dist/latest-v12.x/docs/api/v8.html#v8_v8_writeheapsnapshot_filename) for file name options
+
+You need to have a way to invoke it without stopping the process, so calling it in a http handler or as a reaction to a signal from the operating system is advised.
```
gireeshpunathil: does it make sense to show example code here? for both http end point and the signal handler? while it will come handy for a good programmer, having it here makes it readily consumable?

documentation/memory/step2/using_heap_snapshot.md
```diff
@@ -6,4 +6,37 @@ can also compare multiple snapshots to see differences over time.
...
+1. Start using the functionality you suspect of leaking memory. It's likely it makes some initial allocations that are not the leaking ones.
+1. Take one heap snapshot.
+1. Continue using the functionality for a while, preferably without running anything else in between.
+1. Take another heap snapshot. The difference between the two should mostly contain what was leaking.
+1. Open Chromium/Chrome dev tools and go to *Memory* tab
+1. Load the older snapshot file first, newer one second ![Screenshot: Load button in tools]
+(TODO) 
+1. Select the newer snapshot and switch mode in a dropdown at the top from *Summary* to *Comparison*.
+1. Look for large positive deltas and explore the references that caused them in the bottom panel.
+
```
gireeshpunathil: should we also spend a few lines about interpreting the view / columns in the dev tools? for example the `own size`, `retained size`, ...

documentation/memory/step3/using_gc_traces.md
gireeshpunathil: can we add a paragraph that explains the trace gc output itself, guiding how to interpret those?

gireeshpunathil:
> What's the policy on screenshots?

I guess we could use those abundantly, as they are highly useful and easy to correlate for the users.

documentation/memory/step2/using_heap_snapshot.md
naugtur: I wasn't sure if it belongs in this document. Can do if it's the desired content.
The exercise covers it, at least partially.

documentation/memory/step2/using_heap_snapshot.md
naugtur: Annotated example would work, but I'm not sure if this is the place to explain the terms. 
I'd expect there's devtools docs or a good post somewhere explaining it better than I could.

naugtur:
@gireeshpunathil I'll add the screenshots then. 
Planning to crop them to avoid system window decorations and unnecessary blank spaces. Stop me if you want consistent image dimensions.

mmarchini:
It's also worth mentioning `--heapsnapshot-signal`, which was introduced on Node.js v12(?)

naugtur:
I'll add the warnings and the flag. 
As for the inspector protocol - IMHO it makes sense to recommend adding the code to the app up-front if possible and provide other options for situations where this is not possible. The article could come with a downloadable script or, better, an npm package with the scripts (as node script/binary and the bash version maybe)

@mmarchini Confirm if you find it enough.

documentation/memory/step2/using_heap_snapshot.md
```diff
@@ -6,4 +6,37 @@ can also compare multiple snapshots to see differences over time.
 
 ## How To
 
-//TODO
+### Get the Heap Snapshot
+
+The simplest way to get a Heap Snapshot is to connect a debugger to your process running locally and go to Profiling, choose to take a memory snapshot.
+
+If you need a snapshot from a working process, like an application running on a server, you should implement getting it using:
```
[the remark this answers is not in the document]
naugtur: `websocat` is unlikely to be installed anywhere and getting it installed in a production deployment is very likely to take way longer than to ship a tiny code change to production.   

I'd suggest a package you can put in as dependency or install/npx it to trigger the heapdump.

vmarchaud:
Not sure if it's relevant there but i've wrote a [tool](https://github.com/vmarchaud/openprofiling-node) to start/retrieve heap profile/snapshot. I [wrote](https://github.com/vmarchaud/openprofiling-node/tree/master/packages/openprofiling-inspector-heap-profiler)  [some](https://github.com/vmarchaud/openprofiling-node/tree/master/packages/openprofiling-inspector-heapsnapshot) explaination about the different advantages/drawbacks of using each method if you want to take inspiration.
There is also the [trace-events](http://nodejs.org/dist/latest-v12.x/docs/api/tracing.html) API that allows you to retrieve event about the V8 GC (see [here](https://github.com/vmarchaud/openprofiling-node/tree/master/packages/openprofiling-inspector-trace-events) for example)

I hope it can be useful to other people !

documentation/memory/step2/using_heap_snapshot.md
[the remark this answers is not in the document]
mmarchini: > websocat is unlikely to be installed anywhere and getting it installed in a production deployment is very likely to take way longer than to ship a tiny code change to production.

Depends on the environment. For example, when dealing with microservice architecture, users can have trouble upgrading all microservices (especially when the number of microservices is big) to have a new Node.js dependency loaded at runtime, but ssh-ing into an instance and installing websocat might be viable.

It's also important to consider that the user taking a heap snapshot might not even be a Node.js developer or they might have no access to change the code of the app (I'm thinking operational engineers dealing with an outage, for example).

I agree `websocat` is not ideal, but so far it's the tool I can think of which deals with those scenarios. Installing `websocat` also won't mess with `node_modules`, which might be good for some environments.

(that's not blocking though, and I can send a PR later to update the inspector protocol section with those notes)

mhdawson:
@naugtur thanks for the work on this.

documentation/memory/step3/using_gc_traces.md
naugtur: It's quite a rabbithole. I'd like to finish this PR and dive into that in the next. There's a lot to know there and I feel like I need to learn some more details before I'm comfortable drawing the line and deciding what goes here as enough of a summary.

naugtur:
I'm back. 

Added one more update and I'm hoping this is enough to get merged.
I'd open another PR after this to continue work on the GC traces doc. 

Also, added a file with the pointer used in screenshots so it's easier to add more consistently.

naugtur:
I'm going to assume one approval is enough.

PR #344, opened 2019-12-15
doc: add abnormal termination step 1

legendecas:
Confirming if the process was exited for proactive invocation of `process.exit`.

The option is available after https://github.com/nodejs/node/pull/30516 has been landed.

### DIFF
--- a/documentation/abnormal_termination/step1/using_exit_stack_traces.md
+++ b/documentation/abnormal_termination/step1/using_exit_stack_traces.md
@@ -36,8 +36,10 @@ the invocation to understand the reasons and prevent it from abnormal exits.
 
 If the process doesn't output a stack trace on exit, we can come to a
 conclusion that the process exited for other reasons, and we
-have to deploy following steps of the Abnormal Termination guides.
+can deploy strategy according to the [abnormal termination guides on common symptoms].
 
 ## Useful Links
 
 - https://nodejs.org/docs/latest/api/cli.html#cli_trace_exit
+
+[abnormal termination guides on common symptoms]: ../README.md#Common_Symptoms
--- a/documentation/abnormal_termination/step1/using_exit_stack_traces.md
+++ b/documentation/abnormal_termination/step1/using_exit_stack_traces.md
@@ -1,1 +1,43 @@
-//TODO
+# Using Exit Stack Traces for Debugging
+
+`--trace-exit` CLI option will print a stack trace on every proactive
+`process.exit` invocation. It is convenient to confirm if the abnormal exit of
+the process was initiated from an invocation of `process.exit` or something
+other crashes.
+
+> Caveat:
+> `--trace-exit` CLI option is available since Node.js v13.5.0.
+
+## How To
+
+There are conditions that we used an third party library that we have trivial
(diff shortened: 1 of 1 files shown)


mhdawson:
We probably need more info in this section as well as we should explain how you can use the stack traces for debugging.

documentation/abnormal_termination/step1/using_exit_stack_traces.md
```diff
@@ -1 +1,15 @@
-//TODO
+# Using `--trace-exit` CLI Option
```
mhdawson:
```suggestion

# Using Exit Stack Traces for Debugging

//TODO
# Using `--trace-exit` CLI Option
```

legendecas:
@mhdawson PTAL again :D

documentation/abnormal_termination/step1/using_exit_stack_traces.md
```diff
@@ -1,15 +1,43 @@
...
+    at /node_modules/some-obscure-library:199:32 # this is the actual location.
+    at Module._compile (internal/modules/cjs/loader.js:1139:30)
+    at Module._extensions..js (internal/modules/cjs/loader.js:1159:10)
+    at Module.load (internal/modules/cjs/loader.js:988:32)
+    at Module._load (internal/modules/cjs/loader.js:896:14)
+    at executeUserEntryPoint (internal/modules/run_main.js:71:12)
 ```
 
+With the knowledge of where the calls of `process.exit` is, we can check around
+the invocation to guess the reasons and prevent it from abnormal exits.
```
mmarchini: I'm not a huge fan of using the word `guess` here. The whole point of using instrumentation is to empower users to make informed decisions instead of guessing why things are not working.

```suggestion
the invocation to understand the reasons and prevent it from abnormal exits.
```

documentation/abnormal_termination/step1/using_exit_stack_traces.md
```diff
@@ -1,15 +1,43 @@
...
+
+Or if we find there is no such prints on process exits, we can come to a
```
mmarchini:
```suggestion
If the process doesn't output a stack trace on exit, we can come to a
```

documentation/abnormal_termination/step1/using_exit_stack_traces.md
```diff
@@ -1,15 +1,43 @@
...
+conclusion that the process may exited for some other obscure reasons, and we
```
mmarchini:
```suggestion
conclusion that the process exited for other reasons, and we
```

documentation/abnormal_termination/step1/using_exit_stack_traces.md
```diff
@@ -1,15 +1,43 @@
...
+have to deploy following steps of the Abnormal Termination guides.
```
mmarchini: > , and we have to deploy following steps of the Abnormal Termination guides.

This is vague, I'd rather point the user to specific guides based on symptoms. Might be worth leaving that last part out or leaving a placeholder so we can replace it in the future when those guides are written.

PR #347, opened 2020-01-13
doc: symptom-diven Abnormal Termination intro

mmarchini:
Reorganizing the Abnormal Termination landing page to be more focused on symptoms than tools. Instead of showing users the available tools, this new format will point users to the appropriate tool guide based on their symptoms.

### DIFF
--- a/documentation/abnormal_termination/README.md
+++ b/documentation/abnormal_termination/README.md
@@ -32,7 +32,7 @@ $ # process exited
 There are three reasons for a process to exit silently: 
 
   1. The process was finished by using `process.exit()`
-  2. The process received a  termination signal
+  2. The process received a termination signal
   3. The process reached the end of the JavaScript execution and there are no async handlers attached
 
 To find out if the process exited because of a `process.exit()` call, follow [Using Exit Stack Traces](step1/using_exit_stack_traces.md). If that doesn't work, check if a process is exiting due to a termination signal by following `PLACEHOLDER FOR GUIDE` or if it's exiting because it doesn't have anything else to execute by following `PLACEHOLDER FOR GUIDE`.
--- a/documentation/abnormal_termination/README.md
+++ b/documentation/abnormal_termination/README.md
@@ -1,16 +1,15 @@
 
 # Abnormal Termination
 
-In this document you can learn about how to debug process crashes.
+In this document you can learn about how to debug process crashes or unexpected process exits.
 
 - [Abnormal Termination](#abnormal-termination)
-  - [The process exits without a useful stack trace](#the-process-exits-without-a-useful-stack-trace)
-    - [Symptoms](#symptoms)
-    - [Debugging](#debugging)
+  - [Introduction](#introduction)
+  - [Common Symptoms](#common-symptoms)
+    - [Process silently exits](#process-silently-exits)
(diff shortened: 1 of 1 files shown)


mmarchini:
Motivation: there is no point sending the user to `--trace-exit` guide if they are getting a stack trace on the terminal, or to send them to `Diagnostics Report` if the process is exiting with status 0. They can use these tools, but it shouldn't be their first stop.

mmarchini:
This also helps us see more clearly blind spots in our instrumentation. For example, we don't have any easy way to determine if a process was finished by an external signal or because there are no async handlers attached.

sam-github:
> to determine if a process was finished by an external signal or because there are no async handlers attached

Isn't that indicated by process termination status?

mmarchini:
> > to determine if a process was finished by an external signal or because there are no async handlers attached

> Isn't that indicated by process termination status?

Interesting, I was not aware that termination by signal would set the exit status to a specific code (although now that I think about it, that makes total sense).

For SIGTERM on Linux, the status code will be 147, or SIGINT it will be 137, and for SIGABRT 134. While users can look at these codes to unterstand if the process was *potentially* terminated, it doesn't show where the signal came from (yes, that might fall under more advanced tools, but I don't think it's a bad idea to send users to external tools if they are useful). OTOH, there is no guarantee that these status code were generated by a signal (something else in the program could have set the exit status).

So, answering your question: the process termination status is an indication of what happened, not why. Maybe we'll reach a conclusion that there are no tools to help users understand why their program is exiting in these scenarios, but a guide to help them investigate (with things to look for) would still be useful.

sam-github:
Its worth checking out https://linux.die.net/man/3/waitpid -- source of a signal isn't something that is known, not to the  receiver, anyhow. System specific tools might exist to trace inter-process communications.

Also, don't be confused by what the shell does (which is to "pretend" that a child exited with `128 + signalnumber` when it didn't actually exit at all) and underlying process behaviour. Shells do their own special things.

Node doesn't get confused, use child process.spawn on a process that calls `process.exit(147)` and again with a process that calls `process.kill(process.pid, 'SIGSEGV')`, and this difference will be clear in the https://nodejs.org/api/child_process.html#child_process_event_exit information: one will have an exit code of 147, the other a signal of 'SIGSEGV'. That node knows this isn't a special characteristic of node, any decent spawn API (go, python, etc.) knows whether a process exited or was signalled.

As for exit when event loop is empty, node knows this (https://nodejs.org/api/process.html#process_event_beforeexit), but it is considered "normal exit" (exit with code `0`), so the caller can't distinguish between that and `process.exit(0)`.

I hope that info might be useful. 

It shouldn't distract from more tooling support, that's always good, and I think this PRs refactor by symptom makes sense.

documentation/abnormal_termination/README.md
```diff
@@ -21,13 +20,37 @@ from the original stack-trace or we need more information like the internal
...
+
+```console
+$ node index.js
+$ # process exited
+```
+
+There are three reasons for a process to exit silently: 
+
+  1. The process was finished by using `process.exit()`
+  2. The process received a  termination signal
```
cjihrig:
```suggestion
  2. The process received a termination signal
```

PR #350, opened 2020-01-25
doc: change champion of async_hooks to @Qard

Qard:
At the Collaborator Summit following Node+JS Interactive 2019 in Montreal, I took over the champion responsibility for async_hooks.

@ofrobots This PR is replacing you as champion in the doc. Would you like to retain champion status or are you fine with me removing your name here?


mhdawson:
I think this has been open long enough and has a enough approvals to land. Please let me know if you think otherwise and I'll revert.

PR #366, opened 2020-03-17
doc: add sec on investigating navite memory leaks

mhdawson:
Signed-off-by: Michael Dawson <michael_dawson@ca.ibm.com>

### DIFF
--- a/documentation/memory/step4/using_native_tools.md
+++ b/documentation/memory/step4/using_native_tools.md
@@ -6,8 +6,8 @@ in the Node.js dependencies (for example V8).
 
 [Valgrind](https://valgrind.org/) is a tool that can be used to instrument an
 application such that you get reports on memory usage, including potential memory
-leaks once the application terminates. Valgrind support Linux as well as some variants
-including MacOSX.
+leaks once the application terminates. Valgrind supports Linux as well as some Unix-like
+variants including macOS.
 
 We recommend using Valgrind to investigate native memory leaks and more
 detailed information on how to use it with Node.js is available in
--- a/documentation/memory/step4/using_native_tools.md
+++ b/documentation/memory/step4/using_native_tools.md
@@ -11,5 +11,4 @@ including MacOSX.
 
 We recommend using Valgrind to investigate native memory leaks and more
 detailed information on how to use it with Node.js is available in
-[investigating_native_memory_lea(k](https://github.com/nodejs/node/blob/master/doc/guides/investigating_native_memory_leak.md)
-
+[Investigating Memory Leaks with valgrind](https://github.com/nodejs/node/blob/master/doc/guides/investigating_native_memory_leak.md)
--- a/documentation/memory/step4/using_native_tools.md
+++ b/documentation/memory/step4/using_native_tools.md
@@ -1,6 +1,6 @@
 # Investigation Native memory leaks
 
-While JavaScript is is a garbage collected language, memory is also allocated natively 
+While JavaScript is a garbage collected language, memory is also allocated natively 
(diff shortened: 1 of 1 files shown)


mhdawson:
FYI - @hekike my action from last Diagnostic meeting.

documentation/memory/step4/using_native_tools.md
```diff
@@ -1 +1,15 @@
...
+in the Node.js dependencies (for example V8).  
+
+[Valgrind](https://valgrind.org/) is a tool that can be used to instrument an
+application such that you get reports on memory usage, including potential memory
+leaks once the application terminates. Valgrind support Linux as well as some variants
+including MacOSX.
+
+We recommend using Valgrind to investigate native memory leaks and more
+detailed information on how to use it with Node.js is available in
+[investigating_native_memory_lea(k](https://github.com/nodejs/node/blob/master/doc/guides/investigating_native_memory_leak.md)
```
richardlau:
```suggestion
[Investigating Memory Leaks with valgrind](https://github.com/nodejs/node/blob/master/doc/guides/investigating_native_memory_leak.md)
```

documentation/memory/step4/using_native_tools.md
```diff
@@ -1 +1,15 @@
...
+# Investigation Native memory leaks
+
+While JavaScript is is a garbage collected language, memory is also allocated natively 
+in C/C++ code. These allocations may occur in the Node.js source code, addons or
+in the Node.js dependencies (for example V8).  
+
+[Valgrind](https://valgrind.org/) is a tool that can be used to instrument an
+application such that you get reports on memory usage, including potential memory
+leaks once the application terminates. Valgrind support Linux as well as some variants
+including MacOSX.
```
richardlau:
```suggestion
leaks once the application terminates. Valgrind support Linux as well as some Unix-like
variants including macOS.
```

documentation/memory/step4/using_native_tools.md
```diff
@@ -1 +1,15 @@
-//TODO
+# Investigation Native memory leaks
+
+While JavaScript is is a garbage collected language, memory is also allocated natively
```
mmarchini:
```suggestion
While JavaScript is a garbage collected language, memory is also allocated natively 
```

documentation/memory/step4/using_native_tools.md
```diff
@@ -1 +1,15 @@
...
+While JavaScript is is a garbage collected language, memory is also allocated natively 
+in C/C++ code. These allocations may occur in the Node.js source code, addons or
+in the Node.js dependencies (for example V8).  
+
+[Valgrind](https://valgrind.org/) is a tool that can be used to instrument an
+application such that you get reports on memory usage, including potential memory
+leaks once the application terminates. Valgrind support Linux as well as some variants
+including MacOSX.
+
+We recommend using Valgrind to investigate native memory leaks and more
```
naugtur: If I were not involved in diagnostics, as a reader, I wouldn't be sure who "We" stands for here. I was trying to avoid using it for that reason when doing technical writing not in my name.  
To decide if we ;) should word these recommendations one way or another, a style guide for technical writing would be necessary, so until this happens, my comment is void.  
Just wanted to bring up a styleguide for technical writing, because I've recently seen one being created by a professional and it seems valuable.

mmarchini: I kinda agree, but I'd rather not make it harder for folks to write these documents for now, since we're already a small number of folks with limited time.

But I see no problem on suggesting a rewrite here if you want.

(Note that "harder" is subjective. For a technical writer it might be easier to write following a styleguide, but for someone with English as second language and not used to these styleguides, it might be an obstacle)

Issue #211, opened 2018-07-06
Diagnostics "Best Practices" Guide?


mike-kaufman:
Something that would be interesting & valuable to the community is if the WG could come up with a set of diagnostics best practices/techniques for production applications.  I'm a little concerned that this is a bit too ambitious, and a little concerned that one person's "best practice" is another's "really dumb thing to do".  But I'm willing to throw this out to see what comes out of it.  :) 

I have a few thoughts about how this could be structured, but would love to hear if anyone else has ideas/suggestions first.

joyeecheung:
It would be great if we can come up with even just a guide with links to good resources and advertise it so people know what to read when they run into different types of problems in production. We always receive bug reports with obscure screenshots of resource usages in the core repo, documentations like that would be a good place to redirect to instead of nodejs/help.

mhdawson:
It would be great if we can put this together

mike-kaufman:
> into different types of problems in production. We always receive bug reports with obscure screenshots of resource usages in the core repo, 

@joyeecheung, can you point me to some examples here?

gireeshpunathil:
I love this idea, support this, and willing to participate / contribute in ways this group needs it and in ways I can.

While I agree that best practices can be subjective, there are elements of diagnostic steps that are impersonal and to the point - an example would be collecting and analysing heapdumps on memory leak.

Regarding the structure, I see we can have different flows such as:
 - organized based on symptoms [ memory, exception, performance, hang, crash ... ]
 - organized based on tools [ tracer, profiler, dumper, inspector, debugger, reporter ... ]
 - organized based on subsystems [ installer, net, fs, child_process, natives, uv, ... ]

I suggest the symptom based categorization as it leads to faster discovery for consumers.

Rgarding the best practice content, again I see few models:
 - dump of complete diagnostic steps on a failing case, with example code
 - screen-shot assisted illustration of diagnostic methodology
 - plain text elaboration of tools and their usage

I suggest the first model as it leads to improved education for consumers.

mike-kaufman:
@joyeecheung, @gireeshpunathil thanks.   Let me strawman an outline and see if this matches what's in your head - feel free to tear it down. :) 

- Production Configuration Best Practices
  - application logging & log management
    - breakouts for different app types
      - server apps
      - serverless
      - desktop
      - IOT 
  -  APMs
  -  Node Internals Tracing 
     -  configuration
     - interpretation
-  Troubleshooting Tasks
[...]

PR #372, opened 2020-04-01
doc: document how to examine `--trace_gc` output

HarshithaKP:


sam-github:
Slightly off topic, perhaps, but is all this information eventually going to end up in https://nodejs.org/en/docs/guides/ ? It'd be a bit more findable there.

gireeshpunathil:
@sam-github - yes, according to https://github.com/nodejs/diagnostics/issues/211#issuecomment-408168295 and the thread beneath it.

documentation/memory/step3/using_gc_traces.md
```diff
@@ -20,4 +20,18 @@ v8.setFlagsFromString('--trace_gc');
...
+```
+[19278:0x5408db0]  44 ms: Scavenge 2.3 (3.0) -> 1.9 (4.0) MB, 1.2 / 0.0 ms  (average mu = 1.000, current mu = 1.000) allocation failure
+
+[23521:0x10268b000]  120 ms: Mark-sweep 100.7 (122.7) -> 100.6 (122.7) MB, 0.15 / 0.0 ms  (average mu = 0.132, current mu = 0.137) deserialize GC in old space requested
+```
+
+This is how to interpret the trace data:
+
+```
+[PID: isolate] < time taken since GC started in ms> : < type/phase of GC > <heap used before GC call in MB> ( < allocated heap before GC call in MB > ) -> < heap used after GC in MB> ( < allocated heap after GC in MB>) <time spent in GC in ms> [ < reason for GC >]
```
mhdawson, line 36: This is hard to read. A graphic where we point to the sections and say what they are might work better.

If we have examples of how to use the traces to identify problems that would also be good.  For example in terms of running out of memory.  If you first see that that the old space size is continually increasing but it takes too long to hit the max, setting the max old space size to a smaller value can help. 

Another one is that if you see that the time spent in GC is continually increasing or is a large portion of the overall that can mean you are short on memory even if you don't OOM.

HarshithaKP: @mhdawson, Modified according to your suggestion. PTAL.

documentation/memory/step3/using_gc_traces.md
```diff
@@ -20,4 +20,80 @@ v8.setFlagsFromString('--trace_gc');
...
+  <tr>
+    <td>(average mu = 0.132, current mu = 0.137) deserialize GC in old space requested</td>
+    <td>Reason for GC</td>
+  </tr>
+</table>
+
+## Examples of diagnosing memory issues with trace option:
+
+A. How to get context of bad allocations using --trace-gc
+  1. Suppose we observe that the old space is ocntinously increasing.
```
Flarna, line 85:
```suggestion
  1. Suppose we observe that the old space is continously increasing.
```

documentation/memory/step3/using_gc_traces.md
```diff
@@ -20,4 +20,80 @@ v8.setFlagsFromString('--trace_gc');
...
+  5. Allow the program to run, hit the out of memory.
+  6. The produced log gives shows the failing context.
+
+B. How to assert whether too many gc are happening or too many gc is causing an overhead
+  1. Review the trace data, specifically around time between consecutive gcs.
+  2. Review the trace data, specifically around time spent in gc.
+  3. If the time between two gc is less than the time spent in gc, the application is sseverely starving.
+  4. If the time between two gc and the time spent in gc are very high, probably the application can use a smaller heap
+  5. If the time between two gc is much greater than the time spent in gc, application is relatively healthy
+  6. While the actual numbers for these metrics change from workload to workload, a reasonable gap between gcs is 20 minutes, and a reasonable gc time is < 100 ms.
```
Flarna, line 98: Not sure if we should mention concrete times here as it heavily depends on the application and the workload.
In special the 20minues seems to be quite long to me.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: @Flarna, thanks. Fixed it.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: Removed the last step.

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+  <tr>
+    <td>(average mu = 0.132, current mu = 0.137) deserialize GC in old space requested</td>
+    <td>Reason for GC</td>
+  </tr>
+</table>
+
+## Examples of diagnosing memory issues with trace option:
+
+A. How to get context of bad allocations using --trace-gc
+  1. Suppose we observe that the old space is ocntinously increasing.
```
mhdawson:
```suggestion
  1. Suppose we observe that the old space is continously increasing.
```

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+  2. But due to heavy gc, the heap roof is not hit, but the process is slow.
+  3. Review the trace data and figure out how much is the total heap before and after the gc.
+  4. Reduce `--max-old-space-size` such that the total heap is closer to the limit.
+  5. Allow the program to run, hit the out of memory.
+  6. The produced log gives shows the failing context.
+
+B. How to assert whether too many gc are happening or too many gc is causing an overhead
```
mhdawson, line 93:
```suggestion
B. How to assert whether too many gcs are happening or too many gc is causing an overhead
```

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+  1. Review the trace data, specifically around time between consecutive gcs.
+  2. Review the trace data, specifically around time spent in gc.
+  3. If the time between two gc is less than the time spent in gc, the application is sseverely starving.
```
mhdawson, line 96:
```suggestion
  3. If the time between two gc is less than the time spent in gc, the application is severely starving.
```

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+  4. If the time between two gc and the time spent in gc are very high, probably the application can use a smaller heap
```
mhdawson, line 97:
```suggestion
  4. If the time between two gcs and the time spent in gc are very high, probably the application can use a smaller heap
```

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+  5. If the time between two gc is much greater than the time spent in gc, application is relatively healthy
```
mhdawson, line 98:
```suggestion
5. If the time between two gcs is much greater than the time spent in gc, application is relatively healthy
```

documentation/memory/step3/using_gc_traces.md
mhdawson: I had the same thought as well.  20 minutes seems to long. Do we have something to back up suggestions for particular numbers?

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+  6. While the actual numbers for these metrics change from workload to workload, a reasonable gap between gcs is 20 minutes, and a reasonable gc time is < 100 ms.
```
mhdawson: I had the same thought as @Flarna as well. 20 minutes seems to long. Do we have something to back up suggestions for particular numbers?

documentation/memory/step3/using_gc_traces.md
```diff
@@ -32,6 +32,68 @@ Obtained traces of garbage collection looks like the following lines.
...
+## Examples of diagnosing memory issues with trace option:
+
+A. How to get context of bad allocations using --trace-gc
+  1. Suppose we observe that the old space is ocntinously increasing.
+  2. But due to heavy gc, the heap roof is not hit, but the process is slow.
+  3. Review the trace data and figure out how much is the total heap before and after the gc.
+  4. Reduce `--max-old-space-size` such that the total heap is closer to the limit.
+  5. Allow the program to run, hit the out of memory.
+  6. The produced log gives shows the failing context.
+
```
mhdawson, line 92: We might add another which is similar except we reduce old space size in order to ensure we actually have an OOM.  If we see the heap is continually increasing but don't OOM for a long time.  Reducing the old space size can help confirm it is an actual leak versus the heap just increasing because there is lots of space and no need to gc or gc aggressively enough.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: No.I don’t have any data / proof points, so I simply removed it.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: @mhdawson, thanks. Fixed it.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: @mhdawson, thanks. Fixed it.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: ditto.

documentation/memory/step3/using_gc_traces.md
HarshithaKP: ditto

documentation/memory/step3/using_gc_traces.md
HarshithaKP: ditto

documentation/memory/step3/using_gc_traces.md
HarshithaKP: @mhdawson, added another example. PTAL.

gireeshpunathil:
ping @mhdawson , @Flarna - looks like your review comments are addressed; could you please have another look?

mmarchini:
Do we want to document a V8 flag that is not public API and we have no control over it being removed or changed?

HarshithaKP:
```
#node --v8-options | grep "trace-gc "
  --trace-gc (print one trace line following each garbage collection)
#
```
it is documented here.

gireeshpunathil:
landed in 4e8b0bcbc210b09945474e9a6570e4da56aa8fca , thanks for the contribution!

PR #378, opened 2020-04-10
doc: document how to use diagnostic report for abnormal termination

HarshithaKP:


documentation/abnormal_termination/step2/using_diagnostic_report.md
```diff
@@ -11,12 +11,69 @@ configuration and component version issues with in your process.
...
 
 1. Open the generated report (either from file or from console)
-2. Take a look to allocation section in the report
-3. Look for the old_space
-    - used is high
-    - available is very low
-4. Look javascriptStack to see the victim of the memory leak which can be the allocation pattern and source to find what causes it (JavaScript stack trace won’t always point to the right frame, can be confusing)
-5. -> see memory related user journeys to identify exact cause
+
+  Report can be generated using the following cli flags: `node --experimental-report --report-on-fatal-error <app.js>`.
```
gireeshpunathil, line 15:
```suggestion
  Report can be generated using the following cli flags: `node --report-on-fatal-error <app.js>`.
```

documentation/abnormal_termination/step2/using_diagnostic_report.md
```diff
@@ -11,12 +11,69 @@ configuration and component version issues with in your process.
 //TODO, raw notes from deep dive:
```
mhdawson, line 11: Should this still be here?

HarshithaKP: @mhdawson, thanks. Removed it.

documentation/abnormal_termination/step2/using_diagnostic_report.md
```diff
@@ -8,15 +8,70 @@ configuration and component version issues with in your process.
...
+      "committedMemory": 5332846624,
+      "capacity": 5332901608,
+      "used": 5332770856,
+      "available": 130752
+  },
+  ```
+ * used is high : Out of allocated  memory 5332910528 bytes, 5332770856 bytes are used.
+ * available is very low : 130752 bytes.
+
+5. Look at `javascriptStack` to see the victim of the memory leak which can be the allocation pattern and /or the actual source. Note that `javaScriptStack` will not always point to the right frame.
```
legendecas, line 43: Maybe `javascriptStack` can be inspected first to get an overview of the cause of the termination?

HarshithaKP: @legendecas, re arranged it. PTAL.

documentation/abnormal_termination/step2/using_diagnostic_report.md
```diff
@@ -8,15 +8,70 @@ configuration and component version issues with in your process.
...
+
+  Report can be generated using the following cli flags: `node --report-on-fatal-error <app.js>`.
+  Report can be generated using the following programatic configuration:
+  ```js
+  process.report...
+  ```
+
+2. Open the generated report which is in the format `report.20200410.114639.4623.0.001.json`. 
+
+3. Look at `javascriptStack` to see the victim of the memory leak which can be the allocation pattern and /or the actual source. Note that `javaScriptStack` will not always point to the right frame.
```
legendecas, line 21: This document is about abnormal termination. So I'm wondering if it's worth documenting about how to determine it is an OOM crash from the diagnostic report? 

Basically, one should first check the abnormal termination is not caused by any programmatic issues like code exceptions by checking the `javascriptStack` to see if there are any clues. 

Then, if there are no clues on `javascriptStack`, we can take a look at the `nativeStack` and `javascriptHeap` to see if it's crashed for OOM.

HarshithaKP: @legendecas, thanks. Made changes accordingly. PTAL.

legendecas (approved):

PR #387, opened 2020-05-06
doc: add minutes for 2020-05-06

mmarchini:


mmarchini:
Action items extracted from the minutes:

  * [ ] @mmarchini will comment on the reportVersion issue suggesting schema instead of versioning
  * [ ] Peter and Gireesh will try to document the process for the User Journeys -> documentation transfer
  * [ ] @mmarchini will push forward the unofficial build with USDT enabled
  * [ ] WG will walk through the user journeys to understand where BPF tooling fits best

No9:
@mmarchini I took an action to confirm the kernel compatibility question 
Also @robin-rpr was looking for the slides should I create a folder in the root of this repo called ebpf and drop them in there?

mmarchini:
Yeah that sounds good. Probably worth doing that in another PR though

PR #408, opened 2020-07-10
chore(github): add stale action

mmarchini:
Add Stale actions to help keep the issue tracker cleaner. Issues
inactive for 90 days will be marked as `stale`, and issues marked as
stale for 14 days will be automatically closed. The label `never stale`
can be used to skip certain issues.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,17 @@
+name: "Close stale issues"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


mmarchini:
The number of days to mark as stale and to close are a suggestion, we can tweak it as we see fit.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open many days with no activity. It will be closed soon unless the stale label is removed or a comment is made.'
+        stale-issue-label: 'stale'
+        exempt-issue-label: 'never stale'
```
gireeshpunathil, line 15: thought I cannot think of anything concrete, can we think about a better single word for this label? like `immortal` , `active`, `never-stale` etc.

mmarchini: Sorry, ended up merging without making this change. I'll create the `never stale` label just so that the action doesn't break, but we can rename later.

PR #411, opened 2020-07-17
doc: add button to join Slack channel

mmarchini:
Add button to join the recently created #diagnostics-wg Slack channel on Node Slackers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
 # Diagnostics Working Group
+
+[![Join us on nodeslackers.com (channel #diagnostics-wg)](https://img.shields.io/badge/Join%20us%20on-nodeslackers.com-blue)](https://node-js.slack.com/archives/C016U31J15M)
+
 The goal of this WG is to ensure Node provides a set of comprehensive, documented, extensible diagnostic protocols, formats, and APIs to enable tool vendors to provide reliable diagnostic tools for Node.
 
 Work is divided into several domains:


README.md
```diff
@@ -1,4 +1,7 @@
 # Diagnostics Working Group
+
+[![Join us on Slack](https://img.shields.io/badge/Join%20us%20on-Slack-e01563.svg)](https://node-js.slack.com/archives/C016U31J15M)
```
mhdawson, line 3:
```suggestion
[![Join us on Slack(#diagnostics-wg)](https://img.shields.io/badge/Join%20us%20on-Slack-e01563.svg)](https://node-js.slack.com/archives/C016U31J15M)
```

mhdawson: The name of the channel is useful if you are already a member, suggest we add it.

Qard:
I think that link would only work for people already _in_ the Slack so might be a good idea to also include a link to http://www.nodeslackers.com.

mmarchini:
Any suggestions on how to include the link?

RafaelGSS:
> Any suggestions on how to include the link?

Maybe a step by step to subscribe on slack (confirmation on email) and gives the channel link after

mmarchini:
Could be, although a badge doesn't fit well in this case. So either we would move the link to https://github.com/nodejs/diagnostics#logistics with the step-by-step (which makes it less visible), or we could keep the info in the badge + in the logistics section (which is redundant). Both have drawbacks, we just need to choose which one is more acceptable for us.

RafaelGSS:
So, maybe change the text to "Join our **channel** slack" and if necessary can go to https://github.com/nodejs/diagnostics#logistics to create an account. What do you think?

mmarchini:
Another suggestion for the label:

[Screen Shot 2020-07-23 at 13 39 13](https://img.shields.io/badge/Join%20us%20on-nodeslackers.com-blue)

PR #420, opened 2020-07-28
Moving Matteo to emeritus

mcollina:
I have not participated in meetings for a while, and my availability for OSS work is going to decrease by the end of the year for personal reasons (if you are curious, hit me up privately! It's a good thing!).

I've removed myself from the diagnostics team already on github.


mcollina:
(I'm happy to keep reviewing PRs on nodejs/node that are async hooks./asyncresource/* related, just ping me directly if you need me to check things out!)

PR #463, opened 2021-01-27
Emeritus Member username updated

marsonya:
Glen Keane seems to have renamed their Github username from 'thekemkid' to 'GlenTiki'.
Github seems to be redirecting (301) the user's repo links but not the profile link.


mhdawson (approved):
LGTM

Issue #454, opened 2020-11-27
Transpose Deep Dive documents to documents/


RafaelGSS:
In some past WG Meetings, @gireeshpunathil told me that would be great to transpose some documents created on Deep Dive Meetings (see #439) to the `documents/` folder at this repository.

This issue is to clarify this action and revisit some files that have been there.

RafaelGSS:
@gireeshpunathil Can you provide a better context here?

gireeshpunathil:
ok.

The purpose is to transform the deep dive document into the structure defined in the documentation folder in the repo https://github.com/nodejs/diagnostics/tree/master/documentation.

The reason why we are doing is, the deev dive content is mostly free flow write-ups came out of discussions in WG sittings, largely for our consumption. That needs to be in a format that is externally consumable.

To know how to do that, let us take an example of https://github.com/nodejs/diagnostics/tree/master/documentation/profiling

 - Readme.md should contain a high level overview of what use case this flow is addressing (done)
 - setup.md may contain pre-requisites for the system to be able to follow diagnostics (not done)
 - investigation_flow.md may contain high level steps on the problem determination (not done)
 - case_study.md may contain an investigation of a specific use case with real data (node done)
 - step1, step2, .. may contain different tools / methods to solve a problem (not done)


`may contain` indicates that those documents are not must, only requires if the specific use case requires one (considered on a case basis)
`different setps` are present only if a problem is solved in different ways (tool1, tool2 etc.)


Hope this helps!

RafaelGSS:
Excellent. 

Probably would be great to check the folders inside `documentation/` and provide a clear vision of what we want to expose at each folder. For instance, the `cpu` and `profiling` can expose the same thing in some way.

1. How we can deal with this cross-context?
2. What do you think about creating a `README` inside `documentation` to explain each topic before dig into it?

I believe that with these clear ways we can create the topics.

gireeshpunathil:
1. I don't think this happens with every use case / user journey. Let me try to inspect the cpu / profiling cross context and get back on this.

2. Yes, a good idea! (go for it if you like)

RafaelGSS:
[...]

PR #499, opened 2021-09-17
update: cleanup profiling root folder

RafaelGSS:
@nodejs/diagnostics 

I was writing the tutorial for `--prof` usage but looks like https://nodejs.org/en/docs/guides/simple-profiling/ did a great job on this context and there is no reason to document it again.

That said, I would like to revive https://github.com/nodejs/diagnostics/issues/454 in order to document only missing tutorials. If there are no objections, I would close this PR and open a new issue to tracking missing tutorials.


mhdawson:
It makes sense to me to avoid duplicating good content that is out there. Helping people get to that content adds value.

mhdawson:
It looks like this would just add those references, so it would still make sense?

RafaelGSS:
> It looks like this would just add those references, so it would still make sense?

I thinking in close this PR and open an issue to track the real missing tutorials. However, I'm fine by just adding a reference in those cases. What do you think?

gireeshpunathil (approved):

Issue #502, opened 2021-09-30
User Journey tracking documentation


RafaelGSS:
This issue aims to track the remaining tutorials under [documentation](https://github.com/nodejs/diagnostics/tree/main/documentation) folder.

The https://github.com/nodejs/diagnostics/issues/439 contains the documents created in the WG Deep Dive and we are working to transpose them into live documents.

- [ ] Abnormal Termination
  - [X] Using exit stack traces
  - [X] Using diagnotics report
  - [ ] Using lldb
  - [X] Using valgrind
- [ ] CPU
- [ ] Crash
  - [ ] Using node report
  - [ ] Using llnode
- [ ] Hang
  - [X] Using trace sigint
  - [X] Using Chrome DevTools
  - [ ] Using node report
- [ ] Live Debugging
  - [ ] Using Node Inspector
- [ ] Memory
  - [ ] Using Heap Profiler
  - [X] Using Heap Snapshot
  - [X] Using GC Traces
  - [X] Using Native Tools
- [ ] Profiling
  - [X] Using V8 Profiler
  - [X] Using Linux Perf
  - [ ] Using Native Tools
- [ ] Tracing
  - [ ] Using Trace Events

Feel free to update this description when a tutorial is done.

cc: @nodejs/diagnostics

RafaelGSS:
What we are supposed to do in _CPU_ journey? It looks pretty similar to what we did in _Profiling_.

mhdawson:
@RafaelGSS a few thoughts:

1) The goal was to be based on a user journey (ie problem->resolution) versus tool based. In that context we should probably rename "Profiling" to "Poor Performance".
2) CPU should then likely be "Excessive CPU usage". The symptoms would be different (ie performance could be fine, but you may be using more CPU than expected/wanted).  The Debugging might then be the same as figuring out where CPU is used will be the same between the 2.

RafaelGSS:
I'm adding it to the agenda to discuss how we can expose it to the community in a call for help. Maybe a tweet could be enough?

tony-go:
Hey @mhdawson @RafaelGSS 👋 

A question regarding "Monitoring Excessive CPU Usage" we want to help Node.Js users to have a real-time CPU monitoring solution or an on-demand tool that is runnable locally?

RafaelGSS:
> Hey @mhdawson @RafaelGSS wave
> 
[...]

PR #504, opened 2021-10-11
docs: add heap profiling tutorial

RafaelGSS:
Address #502


mhdawson:
@RafaelGSS thanks for working on this>. I think we should also cover generating a heap dump with https://nodejs.org/api/v8.html#v8_v8_getheapsnapshot and the related methods.

mhdawson:
@RafaelGSS thanks for clarifying.  I think we could use an intro under the top level heading to explain what this doc is covering and introduce the top options that will be covered.

PR #525, opened 2022-02-04
docs: update valgrind investigation

RafaelGSS:
It looks like this guide is no longer in the Node.js documentation. So, I've found it in the gitmemory and updated step-4.


richardlau:
> It looks like this guide is no longer in the Node.js documentation. 

It was moved by @mhdawson as part of https://github.com/nodejs/node/pull/41408. It's more of a guide than the other things that were moved, so perhaps it should be moved back to docs/guides in the main repo?

RafaelGSS:
> > It looks like this guide is no longer in the Node.js documentation.
> 
> It was moved by @mhdawson as part of [nodejs/node#41408](https://github.com/nodejs/node/pull/41408). It's more of a guide than the other things that were moved, so perhaps it should be moved back to docs/guides in the main repo?

Well, I'm fine in both locations, the idea of this repository is that in the future it will be merged in the main repo (AFAIK)

mhdawson:
I'd suggest we just leave it where it is and update the link. We can figure out if we want to move the doc when ` the idea of this repository is that in the future it will be merged in the main repo ` comes to pass.

richardlau (approved):

gireeshpunathil (approved):

PR #572, opened 2022-07-18
doc: revamp gc trace documentation

tony-go:
Hi Diag Fam 👋 

## Context

I recently wrote down a [guide](https://github.com/nodejs/nodejs.org/pull/4590/files) and pushed a PR on the nodejs.org repo. 

As discussed in the recent meeting, it was probably not the best approach as we already have a guide in this repo. Also, the main strategy was to write documentation here first, then move it to the nodejs.org repo as a guide.

## Update

**EDIT**: I finally just keep exercises here. The content is available in this PR: https://github.com/nodejs/nodejs.org/pull/4775


documentation/memory/step3/using_gc_traces.md
```diff
@@ -1,86 +1,229 @@
...
+* Interpret traces
+* Find memory leak source
 
-## Runnig with garbage collection traces
-You can see traces for garbage collection in console output of your process using the `--trace_gc` flag.
+There's probably a lot of stuff to learn about how the garbage collector works, but if you have to know one thing, it's that when GC is running, your code is not.
 
+You may want to know how often and long the garbage collection runs.
+
+## Setup
```
RafaelGSS, line 14: I'd move it to the separated `.md` file. Maybe `exercise.md` ?

RafaelGSS (commented):
Amazing! I'd just move the changes in `using_gc_traces.md` to a separate document. Since we have merged that documentation to nodejs.org, we need to remove it from here.

documentation/memory/step3/using_gc_traces.md
tony-go: (Sorry for the wait 🙈)

My attention here is to provide a similar experience to what we got in a tutorial. Moving it to a separate file will create overhead and degrade the reading experience.

RafaelGSS: My concern is that it would need to be merged into the Node.js Guides along with the exercises package and I don't think it will work. My suggestion is to move it to a separate document and in the principal one (the one that gets merged recently in the nodejs guides) you include a mention to this exercise -- pointing to the diagnostics repository. What do you think?

tony-go: My idea was:
- in the guide
  - let the full readme with examples
  - add a link to the diag repository (to the `exercise` folder)
- in the diag repository
  - keep the `exercise` folder only

Sounds reasonable?

documentation/memory/step3/exercice/server.js
```diff
@@ -0,0 +1,52 @@
+// node.js dependencies
```
RafaelGSS, line 1: I'd change this exercise to a script instead of an HTTP Server. It will be more reasonable for newcomers.

documentation/memory/step3/using_gc_traces.md
```diff
@@ -1,86 +1,229 @@
 # Tracing garbage collection
 
-There's a lot to learn about how the garbage collector works, but if you learn one thing it's that when GC is running, your code is not.
+This guide will go through the fundamentals of garbage collection traces.
 
-You may want to know how often and how long the garbage collection is running.
+By the end of this guide, you'll be able to:
+* Enable traces in your Node.js application
+* Interpret traces
+* Find memory leak source
```
RafaelGSS, line 8:
```suggestion
* Interpret traces
* Identify potential memory issues in your Node.js application
```

Rarely, one can find a memory leak by just using `--trace-gc` (in real-world applications)

documentation/memory/step3/using_gc_traces.md
RafaelGSS: Ok. I'll review the document as a set of examples then

documentation/memory/step3/using_gc_traces.md
```diff
@@ -1,86 +1,229 @@
...
 
-## Runnig with garbage collection traces
-You can see traces for garbage collection in console output of your process using the `--trace_gc` flag.
+There's probably a lot of stuff to learn about how the garbage collector works, but if you have to know one thing, it's that when GC is running, your code is not.
 
+You may want to know how often and long the garbage collection runs.
+
+## Setup
+
+For the proposal of this guide, we will use a simple web server empowered with [Fastify](https://www.fastify.io/).
```
RafaelGSS, line 16: As mentioned before, I strongly recommend using a minimal application. It will decrease the probability of a further change in this document. For instance, `fastify` can land a release that changes the output of this tracing and the user won't be able to reproduce it.

documentation/memory/step3/using_gc_traces.md
```diff
@@ -1,86 +1,229 @@
...
+    process.exit(1);
+  }
+  console.log(`server listening on ${address}`);
+});
 ```
 
+> Even if the leak is evident here, finding the source of a leak could be cumbersome in the context of a real-world application.
+
+## Running with garbage collection traces
+You can see traces for garbage collection in the console output of your process using the `--trace_gc` flag.
```
RafaelGSS, line 66:
```suggestion

You can see traces for garbage collection in the standard output of your process using the `--trace_gc` flag.
```

documentation/memory/step3/using_gc_traces.md
```diff
@@ -1,86 +1,229 @@
...
+[13973:0x110008000]      271 ms: Scavenge 7.5 (10.5) -> 6.4 (15.8) MB, 0.9 / 0.0 ms  (average mu = 1.000, current mu = 1.000) allocation failure
+server listening on http://127.0.0.1:9999
+```
+
+Hard to read? Maybe we should pass in review a few concepts and explain the outputs of the `--trace-gc` flag.
+
 ### Examining a trace with `--trace_gc`
 
-Obtained traces of garbage collection looks like the following lines.
+The `--trace-gc` flag outputs all garbage collection events in the console. First, let me describe the composition of each line. We'll use the following line as a model:
```
RafaelGSS, line 88:
```suggestion
The `--trace-gc` flag outputs all garbage collection events in the console. The composition of each line can be described as:
```

RafaelGSS: I'd avoid first person communication

documentation/memory/step3/using_gc_traces.md
```diff
@@ -1,86 +1,229 @@
...
+| 2.0                                                         | Heap used after GC in MB                 |
+| (4.2)                                                       | Total heap after GC in MB                |
+| 0.5 / 0.0 ms (average mu = 1.000, current mu = 1.000)       | Time spent in GC in ms                   |
+| allocation failure                                          | Reason for GC                            |
+
+We'll only focus on two events here:
+* Scavenge
+* Mark-sweep
+
+The heap is divided into "spaces." Amongst these, we have a space called the "new" space and another one called the "old" space.
```
RafaelGSS, line 111:
```suggestion
The heap is divided into _spaces_. Amongst these, we have a space called the "new" space and another one called the "old" space.
```

documentation/memory/step3/using_gc_traces.md
tony-go: >  It will decrease the probability of a further change in this document.

What if we use a native HTTP server?

> For instance, fastify can land a release that changes the output of this tracing...

The leak we tend to catch here is related to a `Set` that continuously grows, so I wonder how a fastify update could impact this? 

Also, If I choose a server it's because it looks like a real-world case, while scripting is often far from that, but maybe it's a bias. Let me know.

RafaelGSS: If you strongly believe the HTTP Server would be better, I will not block it. I'm just raising my point of view as a user as well.

tony-go:
Hey @RafaelGSS 👋 I just added a script idea (alternative). Could you please share your thoughts on it?

documentation/memory/step3/exercice/script.js
```diff
@@ -0,0 +1,45 @@
+// node.js dependecies
```
RafaelGSS, line 1:
```suggestion
```

documentation/memory/step3/exercice/script.js
```diff
@@ -0,0 +1,45 @@
...
+import os from 'os';
+import { createHmac } from 'crypto';
+import fs from 'fs/promises';
+
+// constants
```
RafaelGSS, line 6:
```suggestion
```

`len` is not a constant

documentation/memory/step3/exercice/script.js
```diff
@@ -0,0 +1,45 @@
...
+let len = 1_000_000;
+const entries = new Set();
+// fix: const fileName = `entries-${Date.now()}`;
```
RafaelGSS, line 9: I don't get the idea of the `fix: *`. In theory, this script should only provide a constant trace report, so the users can analyse and fix it by themselves. Well, you certainly can provide the solution, but you should be careful...

```js
const entries = new Set();
// fix: const fileName = `entries-${Date.now()}`;
``` 
It seems the `new Set()` is a mistake, but its usage is completely fine. I'd mention the `fix:` just in the `entries.add()`. Something like:

```js
  entries.add(entry); // Danger: this will store in the memory: sizeof(entry) * len
```

documentation/memory/step3/exercice/script.js
```diff
@@ -0,0 +1,45 @@
...
+// execution
+(async () => {
+  await fs.writeFile(fileName, "start");
+  while (len > 0) {
+    await addEntry();
+    process.stdout.write(`~~> ${len} entries to record\r`);
+    len--;
+  };
+
+  setImmediate(summary);
```
RafaelGSS, line 43:
```suggestion
  setImmediate(summary);
```

Why is `setImmediate` needed here?

RafaelGSS (commented):
Sorry to be very strict in this PR :sweat_smile:

documentation/memory/step3/exercice/script.js
tony-go: `fix:` comment is just a memo for me, I'll delete them.

My idea is:
- have a script provide a constant trace report
- explain logs in the post
- have a second script without a memory leak, using fs for example.
- readers can observe the difference in the trace report

WDYT?

----------------------------

Also, if I understand your concern correctly, I should remind to readers that the usage of `Set` it's not a bad practice itself? It's more the way we use it that can push pressure on memory.

documentation/memory/step3/exercice/script.js
tony-go: > Sorry to be very strict in this PR 😅

No biggy ^^ It's a really good exercise to define our writing practice 💪

documentation/memory/step3/exercice/script.js
RafaelGSS: > 

> `fix:` comment is just a memo for me, I'll delete them.
> 
> My idea is:
> 
> * have a script provide a constant trace report
> * explain logs in the post
> * have a second script without a memory leak, using fs for example.
> * readers can observe the difference in the trace report
> 
> WDYT?

:+1: I like it.

> Also, if I understand your concern correctly, I should remind to readers that the usage of Set it's not a bad practice itself? It's more the way we use it that can push pressure on memory.

Yes, that's correct

tony-go:
I integrated the new script into the tutorial @RafaelGSS 

Once you'll finish the review. Should I raise a PR on the node repo?

RafaelGSS (commented):
>  I'd just move the changes in using_gc_traces.md to a separate document

Well, I still support this opinion... but, I won't block it anymore.

cc: @nodejs/diagnostics 

---

I believe there is a typo in `exercice`, that should be `exerciSe`.

---

> Should I raise a PR on the node repo?

Yes, please

tony-go:
> I'd just move the changes in using_gc_traces.md to a separate document

Maybe I didn't get the purpose properly: "Since we have merged that documentation to nodejs.org, we need to remove it from here." - Could you elaborate a bit?

RafaelGSS:
My idea is:

- Create an `README.md` inside `exercise` folder
- The file will contain all the steps to perform the exercise
- So, instead of modifying all the existing content inside `using_gc_traces.md`, you could just include something like:
```diff
+ The Diagnostics WG created an exercise that you can practice and improve your gc-tracing skills, check it _here_ (link to the exercise folder)
```

This way we'll have two kinds of documents, the first one a pure and raw tutorial (using_gc_traces.md) and the second one for beginners (`exercise/README.md`). Does that make sense to you?

tony-go:
> This way we'll have two kinds of documents, the first one a pure and raw tutorial (using_gc_traces.md) and the second one for beginners (`exercise/README.md`). Does that make sense to you?

I understand your point but don't understand why the guide should stay _pure and raw_. When I look at other guides, they seem guided and detailed. Even If I understand the purpose of "pure and raw" I think that is more suited to documentation rather than guides.

Anyway, I could adopt a different approach if we decide to follow that way.

Maybe the team could share his view on this cc @nodejs/diagnostics

documentation/memory/step3/using_gc_traces.md
```diff
@@ -127,63 +332,14 @@ PerformanceEntry {
...
-  </tr>
-  <tr>
-    <td>flags</td>
-    <td>Additional information about garbage collection operation.</td>
-  </tr>
-</table>
-
-For more information, you can refer to [the documentation about performance hooks](https://nodejs.org/api/perf_hooks.html).
-
-## Examples of diagnosing memory issues with trace option:
```
RafaelGSS, line 164: Couldn't we use it somehow? The content is pretty good.

RafaelGSS (approved):
That's fair enough

 So, my final review is just:

- Fix the typo `exercise`
- Perhaps include the `## Examples of diagnosing memory issues with trace option:`

Great work!

RafaelGSS:
Please remember to open your changes in the nodejs.org project. The `using_gc_traces.md` file has been removed from this repository.

documentation/memory/step3/using_gc_traces.md
tony-go: I use them in the body of the guide.

RafaelGSS:
> Seems to be ready @RafaelGSS
> 
> Maybe should we add a README in the exercice folder ?

Yes, please

RafaelGSS (approved):
LGTM.

tony-go:
🎉

PR #587, opened 2022-08-24
doc(llnode): add user journey guide

tony-go:
## Context

This pull request aims to document `llnode` user journey.

Ref: https://github.com/nodejs/llnode/issues/401
Ref: https://github.com/nodejs/diagnostics/issues/502

## Update

- [x] Create a structure
- [x] Validate the structure together
- [ ] Write down the content
- [ ] Reviews the whole
- [ ] Announce & celebrate 🎉

### DIFF
--- a/documentation/crash/step2/using_llnode.md
+++ b/documentation/crash/step2/using_llnode.md
@@ -1,6 +1,9 @@
 # Using `llnode` for post-mortem debugging
 
-You probably already experienced bugs that only happen in production. You'd like to reproduce it and step through instructions, but it will imply stopping your Node.js process and certainly degrade the user experience.
+You probably already experienced crashes that only happen in production.
+You'd like to reproduce it and step through instructions, 
+but it will imply stopping your Node.js process and certainly 
+degrade the user experience.
 
 The purpose of this guide is to help you to find route causes
 of your application crashes in production.
@@ -10,7 +13,9 @@ that will helps you to go thoroughly through this tutorial.
 
 ## Concepts
 
-### What is post-mortem debugging?
+> ✋ If you familliar with the concept of core dump and `lldb`, 
+> please skip this chapter, and start [here](#explore-your-firstcore-dump)
+
 ### What is a core dump?
 ### What is `lldb`?
 
--- a/documentation/crash/step2/using_llnode.md
+++ b/documentation/crash/step2/using_llnode.md
@@ -1,5 +1,13 @@
 # Using `llnode` for post-mortem debugging
 
+You probably already experienced bugs that only happen in production. You'd like to reproduce it and step through instructions, but it will imply stopping your Node.js process and certainly degrade the user experience.
+
+The purpose of this guide is to help you to find route causes
+of your application crashes in production.
+
+Before jumping into the code, let's clarify a few core concepts
+that will helps you to go thoroughly through this tutorial.
+
 ## Concepts
 
 ### What is post-mortem debugging?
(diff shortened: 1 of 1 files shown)


tony-go:
Why this structure?

My idea was to slowly bring the reader from basic theoretical concepts to basic usage of `lldb`. Then, once the reader sees the missing parts, we call `llnode` to the rescue and learn basic usage.

WDYT?

No9:
Love it!! I am a big fan of this exploratory style of tutorial. 
Especially the "it's not doing what was expected this is how we fix it".

If I could suggest a finesse: 
Do you think it would be worth while continuing with that concept into the third section? 
Specifically having the solution to the crash being a variable that is only available in the objects and not in the backtrace.
This "extra lightbulb" would demonstrate the value of having the dump over a just a backtrace which is what most people seem to think a core dump is for.

RafaelGSS (commented):
The diagnostics documentation is around the User Journey. We normally try to simulate a real-world use case, for example: https://github.com/nodejs/diagnostics/tree/main/documentation/poor-performance#my-application-has-a-poor-performance and then use the steps to fix that.

So, overall the content seems pretty good. But, I believe we'll need to elaborate on the user journey as well.

tony-go:
Thanks for your feedback @No9 & @RafaelGSS 🙌 

> Specifically having the solution to the crash being a variable that is only available in the objects and not in the backtrace. (@No9)

Could you give me an example? Something like an OOM? Where the size of an object is causing the crash.

> The diagnostics documentation is around the User Journey (@RafaelGSS)

The journey we decide to follow here is the "post-mortem debugging," and `llnode` will be the solution. Make sense? Or I didn't get your point?

The journey through these events:

- I have an application that crash, and it's hard to analyze the root causes
- I enable core dump 
- I use `lldb`, but it misses something
- I used `llnode`; I got "all" what I need

No9:
My idea is line with what @RafaelGSS suggested. 
The journey should define a common use case (something that is likely to happen in the real world) to highlight the benefits /features of llnode. 
For Example in this article https://developer.ibm.com/articles/explore-nodejs-core-dumps-using-the-llnode-plugin-for-lldb/ the example steps through a missing function call in a http server.
Not only does it generate the crash but it also demonstrates getting additional information from the dump that you wouldn't get in the backtrace.

tony-go:
Thanks again @No9 :) 

I have exactly the same vision, it's also what I try to accomplish in that one https://nodejs.org/en/docs/guides/diagnostics/memory/using-gc-traces/

Here is what we could do to move forward:
- I'll write down the introduction and fill subparts with a few lines.
- Then you'll probably have a better idea about where I'd like to go

WDYT?

tony-go:
Hey @No9 @RafaelGSS 👋 - I added a few lines and updated the structure. Let me know if it fits your expectations 🙌

documentation/crash/step2/using_llnode.md
```diff
@@ -1 +1,52 @@
...
+
+The purpose of this guide is to help you to find route causes
+of your application crashes in production.
+
+Before jumping into the code, let's clarify a few core concepts
+that will helps you to go thoroughly through this tutorial.
+
+## Concepts
+
+### What is post-mortem debugging?
```
RafaelGSS, line 13: I don't think this is needed. Even being important, I don't see beginners going to this document as a first reference. I see this document as a tutorial on a specific workflow, in this case, using llnode to diagnose application _crash_.

So, for instance, I'd replace the sections of `What is ***` to `Finding function using llnode` or `Get crash report using llnode`

Does that make sense?

tony-go: > I don't think this is needed. Even being important, I don't see beginners going to this document as a first reference. I see this document as a tutorial on a specific workflow, in this case, using llnode to diagnose application crash.

I agree to delete the `What is post-mortem debugging`; it seems too broad. But for the two others (core dumps & lldb) it could be relevant IMO. What we could do is invite people that are familiar with these concepts to go to the next chapter. WDYT?

documentation/crash/step2/using_llnode.md
```diff
@@ -1 +1,57 @@
...
+The purpose of this guide is to help you to find route causes
+of your application crashes in production.
+
+Before jumping into the code, let's clarify a few core concepts
+that will helps you to go thoroughly through this tutorial.
+
+## Concepts
+
+> ✋ If you familliar with the concept of core dump and `lldb`, 
+> please skip this chapter, and start [here](#explore-your-firstcore-dump)
```
legendecas, line 17:
```suggestion
> please skip this chapter, and start [here](#explore-your-first-core-dump)
```

documentation/crash/step2/using_llnode.md
RafaelGSS: SGTM

documentation/crash/step2/using_llnode.md
tony-go: Nice catch

